PR #2, opened 2025-06-18
Workflows and steps v1

maxdml:


dbos/system_database.go
```diff
@@ -493,3 +486,210 @@ func (s *systemDatabase) ListWorkflows(ctx context.Context, input ListWorkflowsD
...
+	ResetDeadline         bool
+	ResetDeduplicationID  bool
+	ResetStartedAtEpochMs bool
+}
+
+type UpdateWorkflowStatusWhere struct {
+	Status *string
+}
+
+func (s *systemDatabase) UpdateWorkflowStatus(ctx context.Context, input UpdateWorkflowStatusDBInput) error {
```
kraftp: This incredibly complex function is going to get us into trouble. We don't need it. Instead of writing a generic "update" function that tries to handle every possible case with a million switches, just do what we do in Python/TS and write the appropriate SQL in the appropriate places, maybe factor out a smaller "update outcome" function for writing outputs and errors. In general, for maintainability, let's track the Python system database implementation as closely as possible.

dbos/system_database.go
```diff
@@ -493,3 +486,210 @@ func (s *systemDatabase) ListWorkflows(ctx context.Context, input ListWorkflowsD
...
+	switch wf.Status {
+	case "SUCCESS", "ERROR", "CANCELLED":
+		// Workflow is already in a terminal state, rollback and return
+		if err := tx.Rollback(ctx); err != nil {
+			return fmt.Errorf("failed to commit transaction: %w", err)
+		}
+		return nil
+	}
+
+	updateInput := UpdateWorkflowStatusDBInput{
```
kraftp: For example, this would be cleaner and more maintainable if it just wrote out the SQL (which is fairly simple) instead of that super-function.

dbos/workflow.go
```diff
@@ -198,6 +211,7 @@ type StepParams struct {
 	BackoffRate int
 }
 
+// XXX do we even need to register steps??
```
kraftp: This is a really good point worth thinking about.

We don't need to register steps, but we do need some wrapper around them that runs the step code. If we don't want to register steps, we could write a "run as step" wrapper function that you'd call to run a function as a step. Maybe that will be a cleaner interface? We'd have to restrict step outputs, but maybe not step inputs. That's similar to what TSv3 does.

dbos/dbos_test.go
```diff
@@ -2,47 +2,91 @@ package dbos
...
+func (w workflowStruct) simpleWorkflowValue(ctx context.Context, input string) (string, error) {
+	return input + "-value", nil
+}
+
+// interface for workflow methods
+type WorkflowInterface interface {
+	Execute(ctx context.Context, input string) (string, error)
+}
+
+type workflowImplementation struct{}
```
qianl15: What's the difference between `workflowImplementation` and `workflowStruct`?
Okay I see `workflowImplementation` is an implementation of `WorkflowInterface`.

Then a more interesting test would be to make sure those workflows can correctly access instance fields.

dbos/workflow.go
```diff
@@ -107,13 +122,14 @@ func runAsWorkflow[P any, R any](ctx context.Context, params WorkflowParams, fn
 	}
 
 	workflowStatus := WorkflowStatus{
-		Status:        "PENDING",
+		Status:        WorkflowStatusPending,
```
qianl15, line 126: If I understand it correctly, in Go we always run workflows in a go routine, so everything is an "asynchronous" call that returns a handle.

dbos/system_database.go
maxdml: agreed

dbos/workflow.go
maxdml: Yes

dbos/dbos_test.go
maxdml: Good point, I'll add that!

I've been testing various ways to wrap functions, just to make sure they work. There's a few cases that probably won't work with recovery, e.g., registering anonymous functions. I'll adapt this suite over time.

dbos/workflow.go
maxdml: We could. Check f26b6eac6136a3256d5a8698deeb8549294fda94 to see what it'd look like.

maxdml: What this means tho is that we lose compile time type validation for the function inputs. At best we can throw an error if they don't match the function signature.

edit: type validation showcased in 63c92035aaac87fbc2b411ae2aa0d8949b65f6c4

kraftp: Can the return type be inferred?

maxdml: Eventually we can also provide both ways: typeless and 1 generic input.

I think we should pick one and see how users respond.

kraftp: Luckily, it's easy to switch between them. Let's start with the registration-time wrapper, make it as simple as we can, and adjust later.

maxdml: > Can the return type be inferred?

Yes the return type can be inferred too

maxdml: Btw, with the 1 generic input `RunAsWorkflow`, I don't think we need registration. We only recover workflows, which is why we have registration for.

PR #3, opened 2025-06-19
base child workflows + idempotency check

maxdml:


dbos/system_database.go
```diff
@@ -216,32 +228,63 @@ func (s *systemDatabase) InsertWorkflowStatus(ctx context.Context, initStatus Wo
...
-	).Scan(
-		&result.Attempts,
-		&result.Status,
-		&result.Name,
-		&result.QueueName,
-		&result.WorkflowDeadlineEpochMs,
-	)
+	var err error
+
+	if input.Tx != nil {
```
kraftp, line 233: When do we call this non-transactionally?

dbos/system_database.go
```diff
@@ -531,6 +622,32 @@ func (s *systemDatabase) CancelWorkflow(ctx context.Context, workflowID string)
...
+			return nil, fmt.Errorf("failed to query workflow status: %w", err)
+		}
+
+		switch status {
+		case WorkflowStatusSuccess:
+			return *outputStr, nil // Assuming outputStr can be directly returned as R
+		case WorkflowStatusError:
+			return nil, errors.New(*errorStr) // Assuming errorStr can be converted to an error type
+		case WorkflowStatusCancelled:
+			return nil, fmt.Errorf("workflow %s was cancelled", workflowID)
```
qianl15, line 645: If the status is `PENDING`, will this code continue as a busy loop here? Because I see `time.Sleep` only applies when there's no row in the table. Maybe we should put a `time.Sleep` after this switch statement too?

dbos/system_database.go
```diff
@@ -298,6 +349,46 @@ func (s *systemDatabase) RecordOperationResult(ctx context.Context, input record
 	return nil
 }
 
+func (s *systemDatabase) RecordChildWorkflow(ctx context.Context, input RecordChildWorkflowDBInput) error {
+	query := `INSERT INTO dbos.operation_outputs
+            (workflow_uuid, function_id, function_name, child_workflow_id)
+            VALUES ($1, $2, $3, $4)
+            ON CONFLICT DO NOTHING`
```
qianl15: I think we don't need `ON CONFLICT DO NOTHING` here, because later we have `// Check for unique constraint violation (conflict ID error)`

See Python implementation https://github.com/dbos-inc/dbos-transact-py/blob/main/dbos/_sys_db.py#L1110-L1132

dbos/system_database.go
maxdml: I don't know 😁

maxdml: I'll probably end up making transaction only -- will get to it when I get to compose it with other functions (e.g. init validation, `send`)

PR #4, opened 2025-06-20
sez/dez inputs/outputs

maxdml:


dbos/system_database.go
```diff
@@ -523,9 +540,37 @@ func (s *systemDatabase) ListWorkflows(ctx context.Context, input ListWorkflowsD
...
+			buf := bytes.NewBuffer(outputBytes)
+			dec := gob.NewDecoder(buf)
+			var output any
+			if err := dec.Decode(&output); err != nil {
+				// If deserialization fails, log the error but continue (don't fail the entire query)
+				fmt.Printf("Warning: failed to decode output for workflow %s: %v\n", wf.ID, err)
+				wf.Output = nil
+			} else {
+				wf.Output = output
+			}
```
maxdml: I think I'll have this fail the entire query actually

dbos/workflow.go
```diff
@@ -121,8 +122,12 @@ type WorkflowParams struct {
 }
 
 func WithWorkflow[P any, R any](fn WorkflowFunc[P, R]) func(ctx context.Context, params WorkflowParams, input P) (WorkflowHandle[R], error) {
-	// Also we should register the wrapped function
 	registerWorkflow(fn)
+	// Registry the input/output types for gob encoding
+	var p P
+	var r R
+	gob.Register(p)
+	gob.Register(r)
```
maxdml, line 134: Transparent to the user -- the `gob` lib must know about the structs before it can deserialize into them

dbos/dbos_test.go
```diff
@@ -414,13 +444,119 @@ func TestWorkflowIdempotency(t *testing.T) {
...
+			t.Fatalf("failed to execute workflow: %v", err)
+		}
+
+		// Block until the workflow completes
+		_, err = handle.GetResult()
+		if err != nil {
+			t.Fatalf("expected no error but got: %v", err)
+		}
+
+		retrieveHandler, err := RetrieveWorkflow[string](handle.GetWorkflowID())
```
maxdml, line 480: Note that we ask the user to declare the type they want for the return value using this method -- the WorkflowPollingHandle is generic.

I think this is good for type safety (we'd have to have typeless handles otherwise) but the programmer must know about the type they expect from their workflow -- which they should.

dbos/system_database.go
kraftp: Careful, since this query is used for observability. In Python we learned the hard way that it has to be resilient to serialization failures, as it's often used for debugging serialization issues.

dbos/dbos_test.go
```diff
@@ -21,6 +21,22 @@ import (
...
+}
+
+type StepInputStruct struct {
+	Data SimpleStruct `json:"data"`
+	ID   string       `json:"id"`
+}
+
+type StepOutputStruct struct {
+	Input    StepInputStruct `json:"input"`
+	StepInfo string          `json:"step_info"`
```
qianl15, line 37: I assume it'll still work without annotating with the json tags `json:"step_info"`?
What are the limitations for input/output structs (I remember you said no channels or functions)? It'd be good to document the expectation somewhere.

dbos/migrations/000001_initial_dbos_schema.up.sql
```diff
@@ -14,7 +14,7 @@ CREATE TABLE dbos.workflow_status (
     authenticated_user TEXT,
     assumed_role TEXT,
     authenticated_roles TEXT,
-    output TEXT,
+    output BYTEA,
```
qianl15: I'm a bit worried about changing the schema here. This will make it hard to maintain across languages. Can we convert the byte blob into a string? Or treat it as a string just to be consistent with other languages?

dbos/dbos_test.go
maxdml: Yes the tags are just for marshalling to JSON. I will remove them wherever not needed.

I can't think of limitations except that functions and channels will *not* be (de)serialized by `encoding/gob`.

We'll document it carefully

dbos/migrations/000001_initial_dbos_schema.up.sql
maxdml: Today we already don't have system DB compatibility across languages (and I don't know if/how we want to support it in the future.)

What would using a TEXT column bring us here? (It should be do-able but only a Golang reader could cast this string again into bytes and deserialize them)

dbos/system_database.go
maxdml: In Python we do raise an error if we can't decode.

We could do this too, but I think setting the columns content to "unable to decode: xxx" would be more useful, so a client can still retrieve a workflow record and get information about the decode error (in addition to error logs)

dbos/migrations/000001_initial_dbos_schema.up.sql
qianl15: System DB code may generate different data (json in TS vs jsonpickle in Python), but I think it's good to keep the schema compatible across languages. I imagine it'll be useful if in the future we want to provide observability tools across languages, or allow different languages to share the same system DB.

dbos/system_database.go
```diff
@@ -390,14 +409,15 @@ func (s *systemDatabase) RecordChildWorkflow(ctx context.Context, input RecordCh
 
 // ListWorkflowsInput represents the input parameters for listing workflows
 type ListWorkflowsDBInput struct {
-	WorkflowName       *string
-	WorkflowIDPrefix   *string
+	WorkflowName       string
```
qianl15, line 412: Curious why you changed it from `*string` to `string`? Usually it's good to use null pointers for undefined values.

dbos/system_database.go
```diff
@@ -390,14 +409,15 @@ func (s *systemDatabase) RecordChildWorkflow(ctx context.Context, input RecordCh
...
+	WorkflowIDPrefix   string
 	WorkflowIDs        []string
-	AuthenticatedUser  *string
+	AuthenticatedUser  string
 	StartTime          time.Time
 	EndTime            time.Time
-	Status             *string
-	ApplicationVersion *string
+	Status             WorkflowStatusType
```
qianl15, line 418: In Python, we changed this to an array. So users can list workflows that have status in an array like `[SUCCESS, PENDING, ...]`

dbos/recovery.go
```diff
@@ -0,0 +1,47 @@
...
+	pendingWorkflows, err := getExecutor().systemDB.ListWorkflows(ctx, ListWorkflowsDBInput{
+		Status:             WorkflowStatusPending,
+		ExecutorIDs:        executorIDs,
+		ApplicationVersion: APP_VERSION,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, workflow := range pendingWorkflows {
```
qianl15, line 21: For workflows with "failed to decode input", should we just skip those?

dbos/system_database.go
maxdml: It is useful only if the zero value of the variable is valid. Otherwise, checking if the variable is set to its zero value is the idiomatic way to check a variable.

For example if you have an integer and 0 is a valid value, well, you need pointer.

dbos/system_database.go
maxdml: Thanks! I'll add that to the next PR 🖖

dbos/recovery.go
maxdml: Yeah good point, we can't start such a workflow...

maxdml: Done.

I am thinking that eventually we might want to move encode/decode in a small utility submodule, like we do in pytho/ts, and move the encode/decode logic outside of the system DB. For now I'm leaving it as is, but let's keep this in mind.

PR #6, opened 2025-06-26
basic queue logic

maxdml:


dbos/queue.go
```diff
@@ -0,0 +1,161 @@
...
+
+// WithWorkerConcurrency sets the worker concurrency for the queue
+func WithWorkerConcurrency(concurrency int) QueueOption {
+	return func(q *workflowQueue) {
+		q.workerConcurrency = &concurrency
+	}
+}
+
+// WithGlobalConcurrency sets the global concurrency for the queue
+func WithGlobalConcurrency(concurrency int) QueueOption {
```
kraftp, line 46: Why do we need these?  Why not just create a parameterized queue by passing parameters into the queue constructor?

dbos/workflow.go
```diff
@@ -176,6 +176,8 @@ type WorkflowParams struct {
 	WorkflowID string
 	Timeout    time.Duration
 	Deadline   time.Time
+	QueueName  string
+	IsEnqueue  bool
```
kraftp: These seem redundant&mdash;why do we need both?

dbos/queue.go
maxdml: This is called "functional option pattern", it's useful when you have a function / object where all the arguments are optional. Otherwise we'd have to pass an empty `QueueParams{}`. (Frankly I'm thinking of using it for the workflow params too given they are often called without any arguments)

dbos/workflow.go
maxdml: Because we use a single function to enqueue and start a workflow (across languages), we have to distinguish whether the workflow must be enqueued or executed.

In both cases the queue name is set. If we want to rely on queue name only, we'd have to do something like set the queue name to `""` in the dequeue thread to mean "this is dequeue".

I think this is more explicit than what we have in the other sdks, namely: `execute_workflow`

kraftp: I don't understand--if you're not enqueueing, what do you pass in as the queue name? It seems very clumsy to have to pass in "", false. Doesn't Go have some notion of optional parameters?

maxdml: > if you're not enqueueing, what do you pass in as the queue name?

If we are executing an enqueued task, we do pass the queue name. IMHO its not strictly necessary because the workflow exists in the database, but that's how we do it in other languages.

Go does not have optional parameters. It has variadic single type parameters. The params are all of different types so having them in a struct makes sense. The other thing we can do is use the functional option pattern (as seen above) to set each parameters. The challenge is that variadic params must be placed at the end of the signature, so it somewhat limits our future ability to support workflow with more than one parameter.

Given that it is very unlikely we'll support more than one input parameter any time soon, I am planning to use the functional option pattern.

dbos/dbos_test.go
```diff
@@ -113,15 +115,15 @@ func blockingStep(ctx context.Context, input string) (string, error) {
 var idempotencyWorkflowWithStepEvent *Event
 
 func idempotencyWorkflowWithStep(ctx context.Context, input string) (int64, error) {
-	RunAsStep(ctx, StepParams{}, idempotencyWorkflow, input)
+	RunAsStep(ctx, idempotencyWorkflow, input)
```
kraftp: Took me a while to figure out why this works&mdash;it's because idempotencyWorkflow isn't actually a workflow.

dbos/queues_test.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+		if res != "test-input" {
+			t.Fatalf("expected workflow result to be 'test-input', got %v", res)
+		}
+	})
+
+	t.Run("EnqueuedWorkflowStartsChildWorkflow", func(t *testing.T) {
+		queueWfWithChild := WithWorkflow(queueWorkflowWithChild)
+
+		handle, err := queueWfWithChild(context.Background(), "test-input", WithQueue(queue.name))
```
kraftp, line 104: I still don't understand this syntax. What is that third argument's type? How do you compose multiple options (ID + queue, for example).

dbos/queues_test.go
```diff
@@ -0,0 +1,246 @@
...
+[x] Normal wf with a step
+[x] enqueued workflow starts a child workflow
+[x] workflow enqueues another workflow
+[x] recover queued workflow
+[] queued workflow times out
+[] scheduled workflow enqueues another workflow
+*/
+
+var (
+	queue                      = NewWorkflowQueue("test-queue")
```
kraftp: How do you set options on a queue? I assume tests for those are next?

dbos/dbos_test.go
maxdml: Thanks for the feedback. I'll keep improving the tests organization.

dbos/queues_test.go
maxdml: Multiple options are added as arguments, for example:

`queueWfWithChild(context.Background(), "test-input", WithQueue(queue.name), WithWorkflowID(id))`

dbos/queues_test.go
maxdml:
```golang
NewWorkflowQueue("name", WithWorkerConcurrency(5), WithGlobalConcurrency(10))
```


yes I will be adding tests for queue concurrency in this PR, before merging

dbos/queue.go
qianl15: From a user experience point of view, I think having an explicit `QueueParams{}` might be better. For example, vscode intellisense can tell you all the options you're able to set. Otherwise, users would need to find all the possible options.

dbos/workflow.go
```diff
@@ -118,50 +118,54 @@ func (h *workflowPollingHandle[R]) GetWorkflowID() string {
...
 	// Registry the input/output types for gob encoding
 	var p P
 	var r R
 	gob.Register(p)
 	gob.Register(r)
 
 	// Wrap the function in a durable workflow
-	wrappedFunction := WorkflowWrapperFunc[P, R](func(ctx context.Context, params WorkflowParams, input P) (WorkflowHandle[R], error) {
-		return runAsWorkflow(ctx, params, fn, input)
+	wrappedFunction := WorkflowWrapperFunc[P, R](func(ctx context.Context, input P, opts ...WorkflowOption) (WorkflowHandle[R], error) {
```
qianl15, line 192: Same comment as the one for queues. We can leave it like this and change it later -- it's a bit hard for me to find all possible workflow options without looking at the doc. Can intellisense find all the WorkflowOption functions?

dbos/queue.go
maxdml: What I like with the functional options pattern:
- It is popular with go SDKs (e.g., [docker](https://pkg.go.dev/github.com/docker/docker/client#NewClientWithOpts), gRPC)
- It let's you handle defaults cleanly (inside the option setter)

The drawbacks are:
- What you found out: Intellisense works once you start typing the `With` prefix. You have to know that this pattern is in use. Here is [an example of what it looks like](https://pkg.go.dev/github.com/docker/docker/client#Opt) in a generated godoc

Screenshot 2025-07-08 at 12 35 49

Ideally we would be able to evaluate what our users like most. The change is simple.

dbos/workflow.go
maxdml: See reply above

PR #7, opened 2025-07-09
Errors

maxdml:
* Add custom errors
* Check conflicting workflow names at runtime
* Forbid dynamic queue & workflow registrations (I prefer that over just a warning)
* Hash application version from registry at `init()` 
* Add a serialization module

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -180,6 +180,17 @@ func TestAppVersion(t *testing.T) {
 	if _, err := hex.DecodeString(APP_VERSION); err != nil {
 		t.Fatalf("APP_VERSION is not a valid hex string: %v", err)
 	}
+
+	// Replace the registry and verify the hash is different
+	registry = make(map[string]TypedErasedWorkflowWrapperFunc)
+
+	WithWorkflow(func(ctx context.Context, input string) (string, error) {
+		return "new-registry-workflow-" + input, nil
+	})
+	hash2 := computeApplicationVersion()
+	if APP_VERSION == hash2 {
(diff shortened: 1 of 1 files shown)


dbos/dbos.go
```diff
@@ -2,31 +2,67 @@ package dbos
...
 	EXECUTOR_ID = os.Getenv("DBOS__VMID")
 	if EXECUTOR_ID == "" {
-		// Generate a default ID or leave empty based on your requirements
 		EXECUTOR_ID = "local"
 		fmt.Printf("DBOS: DBOS__VMID not set, using default: %s\n", EXECUTOR_ID)
 	}
 
-	fmt.Printf("DBOS: Initialized with APP_VERSION=%s, EXECUTOR_ID=%s\n", APP_VERSION, EXECUTOR_ID)
+	APP_ID = os.Getenv("DBOS__APPID")
+	fmt.Printf("DBOS: Initialized with APP_VERSION=%s, EXECUTOR_ID=%s, APP_ID=%s\n", APP_VERSION, EXECUTOR_ID, APP_ID)
```
kraftp: I wouldn't print app ID, it's a cloud-only thing.

dbos/dbos.go
```diff
@@ -2,31 +2,67 @@ package dbos
...
+	"sort"
 )
 
 var (
 	APP_VERSION string
 	EXECUTOR_ID string
+	APP_ID      string
 )
 
+func computeApplicationVersion() string {
```
kraftp, line 20: We'll also need a lot of tests around versions, especially with how they interact with queues. Can add them later.

PR #8, opened 2025-07-09
Record child get result

maxdml:
Also simplify workflow outcome management (single channel)

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -3,6 +3,7 @@ package dbos
 import (
 	"context"
 	"encoding/gob"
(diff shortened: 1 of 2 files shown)


dbos/system_database.go
```diff
@@ -280,130 +281,6 @@ func (s *systemDatabase) InsertWorkflowStatus(ctx context.Context, input InsertW
 	return &result, nil
 }
 
-type recordOperationResultDBInput struct {
```
maxdml, line 283: this just got moved around a bit

PR #11, opened 2025-07-10
Workflow DLQ

maxdml:
Statically configure max retries during workflow registration.

User exp:
* Provide nothing --> default value
* Set a negative value --> infinite retries
* Provide a positive integer --> max retries

This UX works well with how Golang zero values work.

### DIFF
--- a/dbos/queues_test.go
+++ b/dbos/queues_test.go
@@ -35,11 +35,12 @@ var (
 	successiveEnqueueQueue         = NewWorkflowQueue("test-successive-enqueue-queue")
 	successiveEnqueueStartEvent    = NewEvent()
 	successiveEnqueueCompleteEvent = NewEvent()
+	successiveEnqueueMaxRetries    = 10
 	successiveEnqueueWorkflow      = WithWorkflow(func(ctx context.Context, input string) (string, error) {
 		successiveEnqueueStartEvent.Set()
 		successiveEnqueueCompleteEvent.Wait()
 		return input, nil
-	})
+	}, WithMaxRetries(successiveEnqueueMaxRetries))
 )
 
 func queueWorkflow(ctx context.Context, input string) (string, error) {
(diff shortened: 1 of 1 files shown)


dbos/workflow.go
```diff
@@ -212,15 +216,42 @@ func registerWorkflow(fqn string, fn TypedErasedWorkflowWrapperFunc) {
...
+)
+
+func WithMaxRetries(maxRetries int) WorkflowRegistrationOption {
+	return func(p *WorkflowRegistrationParams) {
+		p.MaxRetries = maxRetries
+	}
 }
 
-func WithWorkflow[P any, R any](fn WorkflowFunc[P, R]) WorkflowWrapperFunc[P, R] {
+func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrationOption) WorkflowWrapperFunc[P, R] {
```
maxdml, line 242: the reason why I preferred the with approach, in this case: there are 2 alternatives, accept a list of parameters

```golang
func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], maxRetries int) WorkflowWrapperFunc[P, R]
```

or a struct

```golang
type Inputs[P any, R any] struct {
 fn WorkflowFunc[P, R]
 maxRetries int
}

func WithWorkflow[P any, R any](inputs Inputs ) WorkflowWrapperFunc[P, R]
```

With the first approach, the UX is bad because the user always needs to pass a value (no default values in golang)

with the second approach, the UX is cumbersome because the compiler cannot infer the generics anymore, and the caller has to do something like

```golang
  simpleWf = WithWorkflow(WorkflowRegistrationInput[string, string]{
        WorkflowFunction: simpleWorkflow,
    })
```

(struct literals generics cannot be inferred, but functions can)

dbos/workflows_test.go
```diff
@@ -679,3 +679,190 @@ func TestWorkflowRecovery(t *testing.T) {
...
+	setupDBOS(t)
+
+	t.Run("DeadLetterQueueBehavior", func(t *testing.T) {
+		deadLetterQueueEvent = NewEvent()
+		deadLetterQueueStartEvent = NewEvent()
+		recoveryCount = 0
+
+		// Start a workflow that blocks forever
+		wfID := uuid.NewString()
+		handle, err := deadLetterQueueWf(context.Background(), "test", WithWorkflowID(wfID))
```
kraftp, line 715: Also check that enqueueing the workflow repeatedly does not cause DLQ issues

maxdml: Thanks, adjusted the test in https://github.com/dbos-inc/dbos-transact-go/pull/11/commits/d2011725374a2b877794ed11e0ddb8f8d73cc818

PR #12, opened 2025-07-10
Scheduled workflows

maxdml:
In a nutshell, https://pkg.go.dev/github.com/robfig/cron does the job of spawning goroutine at desired intervals.

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -279,7 +279,7 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
 
 	// If this is a scheduled workflow, register a cron job
 	if registrationParams.CronSchedule != "" {
-		if reflect.TypeOf(p) != reflect.TypeOf(ScheduledWorkflowInput{}) {
+		if reflect.TypeOf(p) != reflect.TypeOf(time.Time{}) {
 			panic(fmt.Sprintf("scheduled workflow function must accept ScheduledWorkflowInput as input, got %T", p))
 		}
 		if workflowScheduler == nil {
@@ -299,11 +299,7 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
 				scheduledTime = entry.Next
 			}
 			wfID := fmt.Sprintf("sched-%s-%s", fqn, scheduledTime) // XXX we can rethink the format
-			input := ScheduledWorkflowInput{
-				StartTime:     time.Now(),
-				ScheduledTime: scheduledTime,
-			}
-			wrappedFunction(context.Background(), any(input).(P), WithWorkflowID(wfID), WithQueue(DBOS_INTERNAL_QUEUE_NAME))
+			wrappedFunction(context.Background(), any(scheduledTime).(P), WithWorkflowID(wfID), WithQueue(DBOS_INTERNAL_QUEUE_NAME))
 		})
 		if err != nil {
 			panic(fmt.Sprintf("failed to register scheduled workflow: %v", err))
@@ -338,11 +334,6 @@ const workflowStateKey contextKey = "workflowState"
 
 type WorkflowFunc[P any, R any] func(ctx context.Context, input P) (R, error)
 type WorkflowWrapperFunc[P any, R any] func(ctx context.Context, input P, opts ...WorkflowOption) (WorkflowHandle[R], error)
-
-type ScheduledWorkflowInput struct {
-	StartTime     time.Time
-	ScheduledTime time.Time
-}
 
 type WorkflowParams struct {
 	WorkflowID         string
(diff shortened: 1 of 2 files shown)


dbos/workflow.go
```diff
@@ -267,8 +277,25 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
...
+		if workflowScheduler == nil {
+			workflowScheduler = cron.New(cron.WithSeconds())
+		}
+		_, err := workflowScheduler.AddFunc(registrationParams.CronSchedule, func() {
+			// Execute the workflow on the cron schedule once DBOS is launched
+			if getExecutor() == nil {
+				return
+			}
+			wfID := fmt.Sprintf("sched-%s-%s", fqn, time.Now())
+			wrappedFunction(context.Background(), p, WithWorkflowID(wfID))
```
kraftp: Scheduled workflows should enqueue

dbos/workflows_test.go
```diff
@@ -866,3 +867,62 @@ func TestWorkflowDeadLetterQueue(t *testing.T) {
 		}
 	})
 }
+
+var (
+	startTime  = time.Now()
+	counter    = 0
+	counter1Ch = make(chan int, 100)
+	_          = WithWorkflow(func(ctx context.Context, input string) (string, error) {
```
kraftp: What is this string input?

maxdml: Updated this to expect a time -- sorry

dbos/workflow.go
```diff
@@ -267,8 +277,29 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
...
+		if workflowScheduler == nil {
+			workflowScheduler = cron.New(cron.WithSeconds())
+		}
+		_, err := workflowScheduler.AddFunc(registrationParams.CronSchedule, func() {
+			// Execute the workflow on the cron schedule once DBOS is launched
+			if getExecutor() == nil {
+				return
+			}
+			startTime := time.Now()
+			wfID := fmt.Sprintf("sched-%s-%s", fqn, startTime)
```
qianl15: One problem with using `startTime` in the workflow ID is that if there are multiple VMs running the same scheduled workflow, then this workflow may actually execute multiple times, because the actual time may differ by a few seconds. That's why we use the scheduled time in TS and Python.

qianl15: How hard would it be to extract the scheduled time from crontab?

maxdml: We'd have to create our own Job implementation that can access the scheduling time.

dbos/workflow.go
maxdml: fixed

dbos/workflow.go
maxdml: Found a very easy way. Implemented. Thanks for the nudge.

dbos/workflow.go
```diff
@@ -267,8 +277,40 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
...
+			// Get the scheduled time from the cron entry
+			entry := workflowScheduler.Entry(entryID)
+			scheduledTime := entry.Prev
+			if scheduledTime.IsZero() {
+				// Fallback to current time if prev is not available (first run)
+				scheduledTime = time.Now()
+			}
+			wfID := fmt.Sprintf("sched-%s-%s", fqn, scheduledTime) // XXX we can rethink the format
+			input := ScheduledWorkflowInput{
+				StartTime:     time.Now(),
```
kraftp: I think we can get rid of start time and only have scheduled. Having both start and scheduled time is in my opinion more confusing than useful, the only reason they're still both in TS/Python is backwards compatibility.

dbos/workflow.go
```diff
@@ -267,8 +277,40 @@ func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...WorkflowRegistrat
...
+		entryID, err := workflowScheduler.AddFunc(registrationParams.CronSchedule, func() {
+			// Execute the workflow on the cron schedule once DBOS is launched
+			if getExecutor() == nil {
+				return
+			}
+			// Get the scheduled time from the cron entry
+			entry := workflowScheduler.Entry(entryID)
+			scheduledTime := entry.Prev
+			if scheduledTime.IsZero() {
+				// Fallback to current time if prev is not available (first run)
```
kraftp: I'd be careful here, this will lead to duplicate workflows in a distributed setting. There's no way to get the actual scheduled time?

maxdml: the right thing to do is to use "Next" when this is zero. For the first run, this should be the scheduled time. I am saying this after reading the code of the package: https://github.com/robfig/cron/blob/e843a09e5b2db454d77aad25b1660173445fb2fc/cron.go#L257

PR #18, opened 2025-07-16
Send recv

maxdml:
First implementation of send/recv. Few things might move around in the future (e.g., were we create the notifier connection). Also this PR does not handle breaks on the connection checking for NOTIFY.

How it works:
- Send/Recv functions are special durable steps
- Recv, before checking in the DB if the message exists, will wait for PG notifications for a set timeout.
- A notification listener loop is created with the system database, and "launched" with it.

Some details:
- Use the builtin PGX `WaitForNotification` and `OnNotification` mechanisms.
- A recv call will create a new channel and share it with the notification listener. Upon receiving a notification, the notification listener will check if a channel was shared for a given payload, and write a value on the channel to signal the receiver.

About what can be sent: to send non built-in types, users must register them with `encoding/gob` because that's what we use to serialize the message. Usual constraints apply (e.g., no exported fields <-> not serializable). Also note that `encoding/gob` cannot encode raw `nil` values.

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1147,7 +1147,23 @@ func (s *systemDatabase) Recv(ctx context.Context, input WorkflowRecvInput) (any
 		return nil, fmt.Errorf("no output recorded in the last recv")
 	}
 
-	// First check if there is already a message available in the database.
+	// First check if there's already a receiver for this workflow/topic to avoid unnecessary database load
+	payload := fmt.Sprintf("%s::%s", destinationID, topic)
+	c := make(chan bool, 1) // Make it buffered to allow the notification listener to post a signal even if the receiver has not reached its select statement yet
+	_, loaded := s.notificationsMap.LoadOrStore(payload, c)
+	if loaded {
+		close(c)
+		fmt.Println("Receive already called for workflow ", destinationID)
+		return nil, NewWorkflowConflictIDError(destinationID)
+	}
+	defer func() {
+		// Clean up the channel after we're done
+		// XXX We should handle panics in this function and make sure we call this. Not a problem for now as panic will crash the importing package.
+		s.notificationsMap.Delete(payload)
+		close(c)
+	}()
+
+	// Now check if there is already a message available in the database.
 	// If not, we'll wait for a notification and timeout
 	var exists bool
 	query := `SELECT EXISTS (SELECT 1 FROM dbos.notifications WHERE destination_uuid = $1 AND topic = $2)`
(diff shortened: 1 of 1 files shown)


dbos/system_database.go
```diff
@@ -905,6 +974,270 @@ func (s *systemDatabase) GetWorkflowSteps(ctx context.Context, workflowID string
...
+        WHERE destination_uuid = (SELECT destination_uuid FROM oldest_entry)
+          AND topic = (SELECT topic FROM oldest_entry)
+          AND created_at_epoch_ms = (SELECT created_at_epoch_ms FROM oldest_entry)
+        RETURNING message`
+
+	var messageString *string
+	err = tx.QueryRow(ctx, query, destinationID, topic).Scan(&messageString)
+	if err != nil {
+		if err == pgx.ErrNoRows {
+			// No message found, record nil result
```
maxdml, line 1206: This should never happen for reasons explained below

dbos/workflows_test.go
```diff
@@ -978,3 +978,272 @@ func TestScheduledWorkflows(t *testing.T) {
...
+	})
+
+	t.Run("RecvTimeout", func(t *testing.T) {
+		// Create a receive workflow that tries to receive a message but no send happens
+		receiveHandle, err := receiveWf(context.Background(), "timeout-test-topic")
+		if err != nil {
+			t.Fatalf("failed to start receive workflow: %v", err)
+		}
+		result, err := receiveHandle.GetResult(context.Background())
+		if result != "--" {
```
kraftp, line 1194: What is --?  Shouldn't it be nil?

qianl15: That's just the artifact of `receiveWorkflow`. It'll concatenate `-` with empty messages...

dbos/system_database.go
```diff
@@ -905,6 +974,270 @@ func (s *systemDatabase) GetWorkflowSteps(ctx context.Context, workflowID string
...
+		return false, fmt.Errorf("failed to check message: %w", err)
+	}
+	if !exists {
+		// We'll do so through registering a channel with the notification listener loop.
+		payload := fmt.Sprintf("%s::%s", destinationID, topic)
+		c := make(chan bool, 1) // Make it buffered to allow the notification listener to post a signal even if the receiver has not reached its select statement yet
+		_, loaded := s.notificationsMap.LoadOrStore(payload, c)
+		if loaded {
+			close(c)
+			fmt.Println("Receive already called for workflow ", destinationID)
```
qianl15: I think this check is better happening before checking the notifications table (see Python https://github.com/dbos-inc/dbos-transact-py/blob/main/dbos/_sys_db.py#L1307)

The reason is to fail faster to avoid more load on the database.

dbos/workflows_test.go
maxdml: We should return a timeout error -- in the case of go, where `nil` are not going to be valid messages, this is easier than in other languages where `undefined` or `None` seem to be valid messages.

dbos/system_database.go
maxdml: fixed

PR #33, opened 2025-07-21
Config system db

maxdml:
Let users provide the system db url programmatically

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"os"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


dbos/queue.go
```diff
@@ -111,17 +111,14 @@ func queueRunner(ctx context.Context) {
 
 	pollingInterval := baseInterval
 
-	// XXX doing this lets the dequeue and the task invokation survive the context cancellation
-	// We might be OK with not doing this. During the tests it results in all sorts of error inside the two functions above due to context cancellation
-	runnerContext := context.WithoutCancel(ctx)
```
maxdml, line 116: This causes a small race condition in the TestWorkerConcurrency test -- I might have to change this test altogether eventually

PR #38, opened 2025-07-24
Send get events

maxdml:
- Implement set/get events
- Modify the notification system to use condition variables: while send/recv work fine with 1 producer: 1 consumer, set/get events allow for concurrent consumers.
- Add a log entry when `dbos.Launch()` recover workflows.
- Fix transaction usage in `Recv()`

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -39,6 +39,8 @@ type SystemDatabase interface {
 	GetWorkflowSteps(ctx context.Context, workflowID string) ([]StepInfo, error)
 	Send(ctx context.Context, input WorkflowSendInput) error
 	Recv(ctx context.Context, input WorkflowRecvInput) (any, error)
+	SetEvent(ctx context.Context, input WorkflowSetEventInput) error
+	GetEvent(ctx context.Context, input WorkflowGetEventInput) (any, error)
 }
 
 type systemDatabase struct {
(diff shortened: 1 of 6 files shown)


dbos/system_database.go
```diff
@@ -1141,42 +1133,54 @@ func (s *systemDatabase) Recv(ctx context.Context, input WorkflowRecvInput) (any
...
+			getLogger().Warn("Recv() timeout reached", "payload", payload, "timeout", input.Timeout)
 		}
 	}
 
 	// Find the oldest message and delete it atomically
+	tx, err := s.pool.Begin(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer tx.Rollback(ctx)
```
maxdml, line 1183: The transaction should start for the sequence:
- consume message
- record Recv() output

But not before

dbos/system_database.go
```diff
@@ -1231,6 +1235,199 @@ func (s *systemDatabase) Recv(ctx context.Context, input WorkflowRecvInput) (any
...
+
+		err = s.RecordOperationResult(ctx, recordInput)
+		if err != nil {
+			return nil, fmt.Errorf("failed to record operation result: %w", err)
+		}
+	}
+
+	return value, nil
+}
+
```
maxdml, line 1432: I trust recording the step outcome + reading the event do *not* need being done transactionally.

dbos/workflows_test.go
```diff
@@ -1085,48 +1110,6 @@ func TestSendRecv(t *testing.T) {
...
-		// sendWorkflow has 3 Send() calls, receiveWorkflow has 3 Recv() calls
-		expectedSendSteps := 3
-		expectedReceiveSteps := 3
-
-		if len(sendSteps) != expectedSendSteps {
-			t.Fatalf("expected %d steps in send workflow, got %d", expectedSendSteps, len(sendSteps))
-		}
-		if len(receiveSteps) != expectedReceiveSteps {
-			t.Fatalf("expected %d steps in receive workflow, got %d", expectedReceiveSteps, len(receiveSteps))
-		}
```
maxdml, line 1129: this was testing the workflow idempotency, not the steps -- already tested elsewhere.

dbos/system_database.go
kraftp: No, they don't, as it doesn't consume the event

PR #42, opened 2025-07-25
Send outside of workflows + send/set message type registration

maxdml:
- Allow standalone `Send()` and `SetEvent()`
- Hash executable for standalone versioning
- Automatically register `time.Time` for `encoding/gob`
- Make `Send` and `SetEvent` generic so we can register the message type with `encoding/gob` automatically

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -40,7 +40,7 @@ type SystemDatabase interface {
 	GetWorkflowSteps(ctx context.Context, workflowID string) ([]StepInfo, error)
 	Send(ctx context.Context, input WorkflowSendInput) error
 	Recv(ctx context.Context, input WorkflowRecvInput) (any, error)
-	SetEvent(ctx context.Context, input WorkflowSetEventInput) error
+	SetEvent(ctx context.Context, input workflowSetEventInputInternal) error
 	GetEvent(ctx context.Context, input WorkflowGetEventInput) (any, error)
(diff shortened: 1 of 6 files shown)


dbos/system_database.go
```diff
@@ -1103,42 +1103,46 @@ func (s *systemDatabase) notificationListenerLoop(ctx context.Context) {
...
-		tx:         tx,
-	}
-	recordedResult, err := s.CheckOperationExecution(ctx, checkInput)
-	if err != nil {
-		return err
-	}
-	if recordedResult != nil {
-		// when hitting this case, recordedResult will be &{<nil> <nil>}
-		return nil
+	// Check if operation was already executed and do nothing if so (only if in workflow)
```
qianl15, line 1136: Are we going to support temp workflows (for exactly-once semantics) in the future? Looks like if send is called outside of a workflow, it doesn't check for anything.

qianl15: For example, it'd be nice to support "here is an idempotency key, make sure we only send a message to a workflow exactly once".

maxdml: It raises the question: do we ever foresee a world where Send() should *not* be idempotent?

qianl15: Hmm I think making it idempotent is nice. The reason is that `send` will simply append a message to the message queue. If send is not idempotent, then the receiver would need to carefully deduplicate.

maxdml: as discussed let's revisit in the future

PR #44, opened 2025-07-30
Toward executor objects

maxdml:
This PR:
- Encapsulate the state of a "DBOS executor" in an interface exported by the package
- Preserves compile-time type checking by exposing package-level methods that accept a DBOS executor interface
- Attempts to keep the programming interface reasonable by creating a `DBOSContext` type which holds both DBOS executor functionalities and extends the native `context.Context` interface.

I want to allow users to replace the implementation of DBOS to fit their use case, most likely testing. For example, consider the widget store checkout workflow taken from this [sibling PR](https://github.com/dbos-inc/dbos-demo-apps/pull/335/).

A user should be able to test, roughly, like this:

```golang
func TestCheckoutWorkflow(t *testing.T) {
         ...
	dbosContextMock := mocks.NewMockDBOSContext(t)
	t.Run("Payment fails", func(t *testing.T) {
		wfID := "test-workflow-id"

		// Set expectations on what DBOS stuff that happens within the workflow
		dbosContextMock.On("GetWorkflowID").Return(wfID, nil)
		dbosContextMock.On("RunAsStep", mock.Anything, "", mock.Anything).Return(1, nil).Once()                                                          // createOrder
		dbosContextMock.On("RunAsStep", mock.Anything, "", mock.Anything).Return(true, nil).Once()                                                       // reserveInventory
		dbosContextMock.On("SetEvent", mock.Anything).Return(nil).Once()                                                                                 // Set payment event
		dbosContextMock.On("Recv", mock.Anything).Return("failed", nil).Once()                                                                           // payment status
		dbosContextMock.On("RunAsStep", mock.Anything, "", mock.Anything).Return("", nil).Once()                                                         // undoReserveInventory
		dbosContextMock.On("RunAsStep", mock.Anything, UpdateOrderStatusInput{OrderID: 1, OrderStatus: CANCELLED}, mock.Anything).Return("", nil).Once() // updateOrderStatus to CANCELLED
		dbosContextMock.On("SetEvent", mock.Anything).Return(nil).Once()                                                                                 // Set workflow event

		res, err := checkoutWorkflow(dbosContextMock, "")
		if err != nil {
			t.Fatalf("checkout workflow failed: %v", err)
		}
		if res != "" {
			t.Fatalf("expected empty result, got %s", res)
		}
		dbosContextMock.AssertExpectations(t)
	})
}
```

To achieve this, while keeping type checking (see the[ checkout workflow here](https://github.com/dbos-inc/dbos-demo-apps/pull/335/files#diff-10a27237178e0f176f8aca5eb10fd6b0318bdd8a040777da6bb76d4d0fc0e712)), this PR does two things:
- For each package-level DBOS method (e.g., `SetEvent`), add a mirror method on the DBOS "executor" interface.
- Mirror methods are "typeless": interface methods cannot be generic unless the interface is, and we don't want a generic executor.

The main issue of this approach is:
- Many methods and data structures considered "internal" to DBOS must now be exposed to the public, because they must provide an implementation of DBOSContext which implements them.
- Our wrapping logic -- decidly an "ungoly" thing to do -- is generic and returns a wrapped workflow which can be injected with a DBOSContext. The issue with this is that if the user wanted to call a wrapped workflow in their test, they'd have to mock the behavior of many internal functions (e.g., UpdateWorkflowOutcome), that they shouldn't have to know about.
- The user sees package-level methods but must mock the interface methods, leading to confusion as to how to configure mocks properly

These things can be mitigated by:
- Asking users to always call their "raw" workflow functions in the test. Not the wrapped ones
- Asking users to test their steps independently, and, in workflows, mocking `RunAsStep`
- Providing a `RunAsWorkflow` function for child workflows

None of these are an issue when users right integration tests w/Postgres. They don't have to mock anything and can create / inject real executors.

At this point I find this approach extremely cumbersome -- the exposition of so many internal methods ruins the user experience and adds immense cognitive load. We are encapsulating state at the sacrifice of users cognitive load.

Likely a better solution can be found by better separation of concerns and using more Go interfaces. Things I am considering:
- Split "executor" and "context" so that contexts can be a generically typed interface.
- Expose a Workflow interface returned at registration, with a `Run` method that can be mocked. (No more "direct function calls tho")

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -259,8 +259,6 @@ func (e *executor) Shutdown() {
 	if logger != nil {
 		logger = nil
 	}
-	// XX now responsiblity of the caller right?
-	e = nil
 }
 
 func GetBinaryHash() (string, error) {
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -69,11 +69,11 @@ type DBOSContext interface {
 	WithValue(key, val any) DBOSContext
 
 	// Workflow registration
-	RegisterWorkflow(fqn string, fn typedErasedWorkflowWrapperFunc, maxRetries int)
-	RegisterScheduledWorkflow(fqn string, fn typedErasedWorkflowWrapperFunc, cronSchedule string, maxRetries int)
+	RegisterWorkflow(fqn string, fn WrappedWorkflowFunc, maxRetries int)
+	RegisterScheduledWorkflow(fqn string, fn WrappedWorkflowFunc, cronSchedule string, maxRetries int)
 
 	// Workflow operations
-	RunAsStep(fn TypeErasedStepFunc, input any, stepName string, opts ...stepOption) (any, error)
+	RunAsStep(fn StepFunc, input any, stepName string, opts ...StepOption) (any, error)
 	Send(input WorkflowSendInputInternal) error
 	Recv(input WorkflowRecvInput) (any, error)
 	SetEvent(input WorkflowSetEventInputInternal) error
@@ -81,14 +81,19 @@ type DBOSContext interface {
 	Sleep(duration time.Duration) (time.Duration, error)
 
 	// Workflow management
-	RetrieveWorkflow(workflowIDs []string) ([]WorkflowStatus, error)
-	CheckChildWorkflow(parentWorkflowID string, stepCounter int) (*string, error)
-	InsertWorkflowStatus(status WorkflowStatus, maxRetries int) (*insertWorkflowResult, error)
-	RecordChildWorkflow(input recordChildWorkflowDBInput) error
+	InsertWorkflowStatus(status WorkflowStatus, maxRetries int) (*InsertWorkflowResult, error)
 	UpdateWorkflowOutcome(workflowID string, status WorkflowStatusType, err error, output any) error
+	RetrieveWorkflow(workflowIDs []string) ([]WorkflowStatus, error)
+	ListWorkflows(input ListWorkflowsDBInput) ([]WorkflowStatus, error)
+	CheckChildWorkflow(parentWorkflowID string, stepCounter int) (*string, error)
+	RecordChildWorkflow(input RecordChildWorkflowDBInput) error
+	CheckOperationExecution(workflowID string, stepID int, stepName string) (*RecordedResult, error)
+	RecordOperationResult(input RecordOperationResultDBInput) error
+	RecordChildGetResult(input RecordChildGetResultDBInput) error
 
 	// Context operations
 	GetWorkflowID() (string, error)
+	AwaitWorkflowResult(workflowID string) (any, error)
 
 	// Accessors
 	GetWorkflowScheduler() *cron.Cron
@@ -191,8 +196,10 @@ func (e *dbosContext) GetWorkflowWg() *sync.WaitGroup {
 
 func NewDBOSContext(inputConfig Config) (DBOSContext, error) {
 	initExecutor := &dbosContext{
-		workflowsWg: &sync.WaitGroup{},
-		ctx:         context.Background(),
+		workflowsWg:      &sync.WaitGroup{},
+		ctx:              context.Background(),
+		workflowRegistry: make(map[string]workflowRegistryEntry),
+		workflowRegMutex: &sync.RWMutex{},
 	}
 
 	// Load & process the configuration
@@ -232,9 +239,6 @@ func NewDBOSContext(inputConfig Config) (DBOSContext, error) {
 	initExecutor.systemDB = systemDB
 	logger.Info("System database initialized")
 
-	// Initialize the workflow registry
-	initExecutor.workflowRegistry = make(map[string]workflowRegistryEntry)
-
 	return initExecutor, nil
 }
 
@@ -255,8 +259,8 @@ func (e *dbosContext) Launch() error {
 	}
 
 	// Create context with cancel function for queue runner
-	// XXX this can now be a cancel function on the executor itself?
-	ctx, cancel := context.WithCancel(context.Background())
+	// FIXME: cancellation now has to go through the DBOSContext
+	ctx, cancel := context.WithCancel(e.GetContext())
 	e.queueRunnerCtx = ctx
 	e.queueRunnerCancelFunc = cancel
 	e.queueRunnerDone = make(chan struct{})
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -19,8 +19,6 @@ var (
 	_DEFAULT_ADMIN_SERVER_PORT = 3001
 )
 
-var workflowScheduler *cron.Cron // Global because accessed during workflow registration before the dbos singleton is initialized
-
 var logger *slog.Logger // Global because accessed everywhere inside the library
 
 func getLogger() *slog.Logger {
@@ -62,25 +60,56 @@ func processConfig(inputConfig *Config) (*Config, error) {
 	return dbosConfig, nil
 }
 
+type DBOSExecutor interface {
+	Launch() error
+	Shutdown()
+
+	RegisterWorkflow(fqn string, fn typedErasedWorkflowWrapperFunc, maxRetries int)
+	RegisterScheduledWorkflow(fqn string, fn typedErasedWorkflowWrapperFunc, cronSchedule string, maxRetries int)
+
+	GetWorkflowScheduler() *cron.Cron
+	GetApplicationVersion() string
+}
+
 var dbos *executor // DBOS singleton instance
 
 type executor struct {
-	systemDB              SystemDatabase
+	systemDB    SystemDatabase
+	adminServer *adminServer
+	config      *Config
+	// Queue runner context and cancel function
 	queueRunnerCtx        context.Context
 	queueRunnerCancelFunc context.CancelFunc
 	queueRunnerDone       chan struct{}
-	adminServer           *adminServer
-	config                *Config
-	applicationVersion    string
-	applicationID         string
-	executorID            string
-	workflowsWg           *sync.WaitGroup
+	// Application metadata
+	applicationVersion string
+	applicationID      string
+	executorID         string
+	// Wait group for workflow goroutines
+	workflowsWg *sync.WaitGroup
+	// Workflow registry
+	workflowRegistry map[string]workflowRegistryEntry
+	workflowRegMutex sync.RWMutex
+	// Workflow scheduler
+	workflowScheduler *cron.Cron
 }
 
-func Initialize(inputConfig Config) error {
+func (e *executor) GetWorkflowScheduler() *cron.Cron {
+	if e.workflowScheduler == nil {
+		e.workflowScheduler = cron.New(cron.WithSeconds())
+	}
+	return e.workflowScheduler
+}
+
+func (e *executor) GetApplicationVersion() string {
+	return e.applicationVersion
+}
+
+// TODO: use a normal builder pattern name (NewDBOSExecutor)
+func Initialize(inputConfig Config) (DBOSExecutor, error) {
 	if dbos != nil {
 		fmt.Println("warning: DBOS instance already initialized, skipping re-initialization")
-		return newInitializationError("DBOS already initialized")
+		return nil, newInitializationError("DBOS already initialized")
 	}
 
 	initExecutor := &executor{
@@ -90,7 +119,7 @@ func Initialize(inputConfig Config) error {
 	// Load & process the configuration
 	config, err := processConfig(&inputConfig)
 	if err != nil {
-		return newInitializationError(err.Error())
+		return nil, newInitializationError(err.Error())
 	}
 	initExecutor.config = config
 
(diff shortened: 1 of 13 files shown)


kraftp (commented):
This is getting better. Really want to see what it looks like in code. One thing that jumps out is that we're now passing name into `RunAsStep` and `RunAsWorkflow`. That's clunky and probably not necessary--even if the type-erased function has the wrong name, we can extract the right name in the wrapper and pass it through the context.

maxdml:
> This is getting better. Really want to see what it looks like in code. One thing that jumps out is that we're now passing name into `RunAsStep` and `RunAsWorkflow`. That's clunky and probably not necessary--even if the type-erased function has the wrong name, we can extract the right name in the wrapper and pass it through the context.

Correct and I've been looking into doing exactly this. However, setting it in the context means using the `Value` method which means the user would.. have to mock it when passing a mocked DBOSContext. And when calling a package-level DBOS function, the user should only have to mock the mirror interface method.

The solution is likely to pass it in the options -- thus allowing user to optionally set workflow and step names, which we planned to do eventually.

This works well for workflows but, won't work with steps if we want to remove functional step options for a revisited step interface accepting a variadic `any` argument. Looking into it :)

kraftp:
> > This is getting better. Really want to see what it looks like in code. One thing that jumps out is that we're now passing name into `RunAsStep` and `RunAsWorkflow`. That's clunky and probably not necessary--even if the type-erased function has the wrong name, we can extract the right name in the wrapper and pass it through the context.
> 
> Correct and I've been looking into doing exactly this. However, setting it in the context means using the `Value` method which means the user would.. have to mock it when passing a mocked `DBOSContext`. And when calling a package-level DBOS function, the user should only have to mock the mirror interface method.
> 
> The solution is likely to pass it in the options -- thus allowing user to optionally set workflow and step names, which we planned to do eventually.
> 
> This works well for workflows but, won't work with steps if we want to remove functional step options for a revisited step interface accepting a variadic `any` argument. Looking into it :)

There has to be a workaround for this--worst case, an in-memory map from type-erased functions to names.

maxdml:
> There has to be a workaround for this--worst case, an in-memory map from type-erased functions to names.

Yup, that's one option, but it won't work when we move all the functional options to the context. Anything the users -- or us, inside the Generic function -- set through the context has to go through values like `ctx.WithStepParams` and has to be mocked by the interface provider

maxdml:
> > > There has to be a workaround for this--worst case, an in-memory map from type-erased functions to names.
> > 
> > 
> > Yup, that's one option, but it won't work when we move all the functional options to the context. Anything the users -- or us, inside the Generic function -- set through the context has to go through values like `ctx.WithStepParams` and has to be mocked by the interface provider
> 
> I'm sure there's a way to circumvent that, but we don't need to figure it out right now--let's get this interface built out, then fix steps, then we'll clean up the details.

Fair enough -- the topics are sufficiently intertwined though that I took the time to find a basic solution that we will improve on. Specifically:
- Expose a `WithValue` method used to create a new `DBOSContext` with a new context value. This is in any case something Go users will expect from a context.
- Use that to allow users to pass in the step name and other step parameters
- In the generic RunAsStep, store the mapping type erased fqn -> user provided fqn
- In the interface RunAsStep, figure out which one to use (user provided step name or the one we stored)

This seems to work well and I can now _almost_ call steps without parameters.

Workflow names are passed through a new functional option.

I did have to fix the way I implemented the Generic RunAsWorkflow. To return a typed handle to the user, this unfortunately as to create a new goroutine to pass in the workflow outcome from an untyped channel to a typed one. Maybe there's way to improve on that too.


Overall the solution is coming together and I am now updating the tests.

dbos/dbos.go
```diff
@@ -119,87 +215,90 @@ func Initialize(inputConfig Config) error {
...
 	// Create context with cancel function for queue runner
-	ctx, cancel := context.WithCancel(context.Background())
-	dbos.queueRunnerCtx = ctx
-	dbos.queueRunnerCancelFunc = cancel
-	dbos.queueRunnerDone = make(chan struct{})
+	// FIXME: cancellation now has to go through the DBOSContext
+	ctx, cancel := context.WithCancel(c.ctx)
+	c.queueRunnerCtx = ctx
+	c.queueRunnerCancelFunc = cancel
+	c.queueRunnerDone = make(chan struct{})
```
maxdml, line 251: In a next PR I'll tighten up the usage of DBOSContext to manage its owned resources

dbos/queue.go
```diff
@@ -73,10 +72,7 @@ func WithMaxTasksPerIteration(maxTasks int) queueOption {
 
 // NewWorkflowQueue creates a new workflow queue with optional configuration
 func NewWorkflowQueue(name string, options ...queueOption) WorkflowQueue {
-	if dbos != nil {
-		getLogger().Warn("NewWorkflowQueue called after DBOS initialization, dynamic registration is not supported")
-		return WorkflowQueue{}
-	}
+	// TODO: Add runtime check for post-initialization registration if needed
```
maxdml, line 75: Will add this in a next PR + prevention of post `Launch` wf registration

dbos/queue.go
```diff
@@ -165,6 +161,7 @@ func queueRunner(ctx context.Context) {
 					}
 				}
 
+				// XXX this demonstrate why contexts cannot be used globally -- the task does not inherit the context used in the program that enqueued it
```
maxdml, line 164: here `ctx` is the root context created by the user.

dbos/system_database.go
```diff
@@ -42,19 +42,19 @@ type SystemDatabase interface {
...
-	SetEvent(ctx context.Context, input workflowSetEventInputInternal) error
+	SetEvent(ctx context.Context, input WorkflowSetEventInput) error
 	GetEvent(ctx context.Context, input WorkflowGetEventInput) (any, error)
 
 	// Timers (special steps)
 	Sleep(ctx context.Context, duration time.Duration) (time.Duration, error)
 
 	// Queues
-	DequeueWorkflows(ctx context.Context, queue WorkflowQueue) ([]dequeuedWorkflow, error)
+	DequeueWorkflows(ctx context.Context, queue WorkflowQueue, executorID, applicationVersion string) ([]dequeuedWorkflow, error)
```
maxdml, line 57: will cleanup inputs (for a struct) in later PRs

dbos/system_database.go
```diff
@@ -644,6 +645,12 @@ func (s *systemDatabase) AwaitWorkflowResult(ctx context.Context, workflowID str
 	query := `SELECT status, output, error FROM dbos.workflow_status WHERE workflow_uuid = $1`
 	var status WorkflowStatusType
 	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
```
maxdml, line 652: `AwaitWorkflowResult` should be cancellable -- work for another PR but placing this here in the meantime.

dbos/system_database.go
```diff
@@ -1115,16 +1122,16 @@ func (s *systemDatabase) notificationListenerLoop(ctx context.Context) {
 
 const _DBOS_NULL_TOPIC = "__null__topic__"
 
-type workflowSendInputInternal struct {
-	destinationID string
-	message       any
-	topic         string
+type WorkflowSendInputInternal struct {
```
maxdml, line 1124: Terrible name that must be made public for the `DBOSContext` to be implementable (`DBOSContext.Send` must take a typeless input)

dbos/utils_test.go
```diff
@@ -54,28 +54,26 @@ func setupDBOS(t *testing.T) {
...
 		AppName:     "test-app",
 	})
 	if err != nil {
 		t.Fatalf("failed to create DBOS instance: %v", err)
 	}
-
-	err = Launch()
-	if err != nil {
-		t.Fatalf("failed to launch DBOS instance: %v", err)
-	}
```
maxdml, line 68: Most tests don't need Launch(). Only useful for queues, recovery thread, and notifications

dbos/workflow.go
```diff
@@ -250,22 +288,28 @@ func WithSchedule(schedule string) workflowRegistrationOption {
...
-func WithWorkflow[P any, R any](fn WorkflowFunc[P, R], opts ...workflowRegistrationOption) WorkflowWrapperFunc[P, R] {
-	if dbos != nil {
-		getLogger().Warn("WithWorkflow called after DBOS initialization, dynamic registration is not supported")
-		return nil
+// RegisterWorkflow registers the provided function as a durable workflow with the provided DBOSContext workflow registry
+// If the workflow is a scheduled workflow (determined by the presence of a cron schedule), it will also register a cron job to execute it
+// RegisterWorkflow is generically typed, allowing us to register the workflow input and output types for gob encoding
+// The registered workflow is wrapped in a typed-erased wrapper which performs runtime type checks and conversions
+// To execute the workflow, use DBOSContext.RunAsWorkflow
+func RegisterWorkflow[P any, R any](ctx DBOSContext, fn GenericWorkflowFunc[P, R], opts ...workflowRegistrationOption) {
```
maxdml, line 296: One drawback of not wrapping anymore and returning the wrapped function, is that we can't automatically inject, at runtime, the registration options (today only max retries), when the user calls `RunAsWorkflow`.

This is very annoying because it means that, if we want the users to *not* have to pass it for every workflow, we must add registration option to the registry and allow RunAsWorkflow, to, at runtime, fetch the registration option (from the registry).

The registry is keyed by the provided workflow name, here the FQN -- maybe a user provided name in the future.

Now, consider that only the package-level RunAsWorkflow nows the FQN. The interface version receives a typeless wrapper that has another name. The registry is attached to the interface, so the package-level object cannot make any assumption about it.

The way we solve this is by adding a private `withWorkflowName` option to the `WorkflowOptions`. The package-level `RunAsWorkflow` injects that option, such that the Interface version of `RunAsWorkflow` can always retrieve its registered name from the parameters and get its register-time options. The wrappers we create during registration (for recovery and scheduled workflows) does the same (they could inject the the max retry option, but this way we have a unified way to get the name from the interface method.)

dbos/workflow.go
```diff
@@ -348,58 +373,134 @@ type workflowParams struct {
...
+		}()
+
+		typedHandle := &workflowHandle[R]{
+			workflowID:  handle.workflowID,
+			outcomeChan: typedOutcomeChan,
+			dbosContext: handle.dbosContext,
+		}
+
+		return typedHandle, nil
+	}
```
maxdml, line 472: We cannot return a generically typed handle directly from the interface `RunAsWorkflow`. This is because the outcome channel itself is typed. This solution has an intermediate goroutine pipe the untyped result from the interface `RunAsWorfklow` to a typed channel, so we can create & return a typed handle to the user. Would love to be suggested alternative ideas.

dbos/workflow.go
```diff
@@ -348,58 +373,134 @@ type workflowParams struct {
...
+	if !exists {
+		return nil, newNonExistentWorkflowError(params.workflowName)
+	}
+	if registeredWorkflow.maxRetries > 0 {
+		params.maxRetries = registeredWorkflow.maxRetries
+	}
 
 	// Check if we are within a workflow (and thus a child workflow)
-	parentWorkflowState, ok := ctx.Value(workflowStateKey).(*workflowState)
+	parentWorkflowState, ok := c.Value(workflowStateKey).(*workflowState)
```
maxdml, line 497: I am still in the process of figuring of whether this will work during child workflows recovery. Because the recovery thread will use the root context, which today doesn't not have a `workflowStateKey`. The recovery thread might be able to reconstruct this, however.

dbos/workflow.go
```diff
@@ -436,219 +537,238 @@ func runAsWorkflow[P any, R any](ctx context.Context, fn WorkflowFunc[P, R], inp
...
 
 /******************************/
 /******* STEP FUNCTIONS *******/
 /******************************/
 
-type StepFunc[P any, R any] func(ctx context.Context, input P) (R, error)
+type StepFunc func(ctx context.Context, input any) (any, error)
+type GenericStepFunc[P any, R any] func(ctx context.Context, input P) (R, error)
+
+const StepParamsKey DBOSContextKey = "stepParams"
```
maxdml, line 642: This is how users can set step parameters, without passing functional options nor a new input parameter.

dbos/workflow.go
```diff
@@ -436,219 +537,238 @@ func runAsWorkflow[P any, R any](ctx context.Context, fn WorkflowFunc[P, R], inp
...
 	}
+
+	return params
 }
 
-// WithMaxInterval sets the maximum delay for retries
-func WithMaxInterval(maxInterval time.Duration) stepOption {
-	return func(p *StepParams) {
-		p.MaxInterval = maxInterval
+var typeErasedStepNameToStepName = make(map[string]string)
```
maxdml, line 682: Again some pain:
- The user might or might not provide a custom step name
- The package level method knows the true name of the function
- The interface method sees only the typed erased wrapper name
- We want package-level and interface methods to have the same signature
- No functional options, no "step params" to smooth the step UX

--> We use an in-memory map to pass the function name to the interface method, which will use it for a default name if the user didn't provide any.

dbos/workflow.go
```diff
@@ -754,20 +888,27 @@ func Recv[R any](ctx context.Context, input WorkflowRecvInput) (R, error) {
 	return typedMessage, nil
 }
 
-type WorkflowSetEventInput[R any] struct {
+type WorkflowSetEventInputGeneric[R any] struct {
```
maxdml, line 889: sadge

dbos/workflows_test.go
```diff
@@ -310,33 +307,33 @@ func stepRetryAlwaysFailsStep(ctx context.Context, input string) (string, error)
...
+		MaxRetries:   5,
+		BaseInterval: 1 * time.Millisecond,
+		MaxInterval:  10 * time.Millisecond,
+	})
 
-var (
-	stepWithinAStepWf = WithWorkflow(stepWithinAStepWorkflow)
-	stepRetryWf       = WithWorkflow(stepRetryWorkflow)
-)
+	return RunAsStep(stepCtx, stepRetryAlwaysFailsStep, input)
```
maxdml, line 323: step params in use

PR #47, opened 2025-08-06
Workflow timeouts

maxdml:
**Design**
This PR brings native Go cancellation facilities to workflows. Users can set a timeout, using `WithTimeout`, to a `DBOSContext` passed to `RunAsWorkflow`.
When the context is cancelled (either due to a timeout or a manual cancellation), DBOS cancellation logic (cancel the workflow in Postgres) will be triggered.
When the workflow runs or is enqueued for the first time, the deadline will be persisted in Postgres.
During recovery, we will recompute the deadline from the value stored in the database and configure the workflow context accordingly.

Details:
- Inside `RunAsWorkflow`, we create an uncancellable context to run DBOS operations and configure the handler. This allows the timeout to apply only to the user function and provide an ergonomic behavior with the handle. (The handle does not "expire", for instance.)
- Because cancellation != preemption, and user functions can run without accounting for the cancellation signal, we make sure to always record their outcome, even if they were cancelled. This does not change their `CANCELLED` status.

Tests:
    [x] workflow timeout
    [x] workflow manual cancellation
    [x] workflow with steps:
        - [x] wait for cancel inside the step triggers workflow cancel
        - [x] shorter deadline can be set for the stepCtx -- parent can handle it as it wishes
        - [x] step can be detached -- workflow still times out
    [x] child workflows
        - [x] child times out: parent gets to handle the error and decide what to do
        - [x] child is detached: status success but parent is cancelled
    [x] queues
        - [x] enqueued workflow times out
        - [x] enqueued workflow that has enqueues another workflow times out (both time out)
        - [x] enqueued workflow enqueues a detached workflow (parent times out, not child)
    [x] recovery

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -655,9 +655,8 @@ func (s *systemDatabase) CancelWorkflow(ctx context.Context, workflowID string) 
 	}
 
 	// Set the workflow's status to CANCELLED and started_at_epoch_ms to NULL (so it does not block the queue, if any)
-	// XXX set dedup id to null? Fix it when we implement deduplication
 	updateStatusQuery := `UPDATE dbos.workflow_status
-						  SET status = $1, updated_at = $2, started_at_epoch_ms = NULL, deduplication_id = NULL
+						  SET status = $1, updated_at = $2, started_at_epoch_ms = NULL
 						  WHERE workflow_uuid = $3`
 
 	_, err = tx.Exec(ctx, updateStatusQuery, WorkflowStatusCancelled, time.Now().UnixMilli(), workflowID)
@@ -1781,7 +1780,7 @@ func (s *systemDatabase) DequeueWorkflows(ctx context.Context, queue WorkflowQue
 	}
 
 	if len(dequeuedIDs) > 0 {
-		// fmt.Printf("[%s] attempting to dequeue %d task(s)\n", queue.Name, len(dequeuedIDs))
+		s.logger.Debug("attempting to dequeue task(s)", "queueName", queue.Name, "numTasks", len(dequeuedIDs))
 	}
 
 	// Update workflows to PENDING status and get their details
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -286,9 +286,10 @@ func (s *systemDatabase) InsertWorkflowStatus(ctx context.Context, input insertW
 		deadline = &millis
 	}
 
-	var timeoutMs int64 = 0
+	var timeoutMs *int64 = nil
 	if input.status.Timeout > 0 {
-		timeoutMs = input.status.Timeout.Milliseconds()
+		millis := input.status.Timeout.Milliseconds()
+		timeoutMs = &millis
 	}
 
 	inputString, err := serialize(input.status.Input)
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -653,10 +653,10 @@ func (s *systemDatabase) CancelWorkflow(ctx context.Context, workflowID string) 
 		return nil
 	}
 
-	// Clear queue-related fields and set the workflow's status to CANCELLED
-	// Note we leave the queue name because it is informative and won't block the queue
+	// Set the workflow's status to CANCELLED and started_at_epoch_ms to NULL (so it does not block the queue, if any)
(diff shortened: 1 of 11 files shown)


dbos/system_database.go
```diff
@@ -630,20 +646,21 @@ func (s *systemDatabase) CancelWorkflow(ctx context.Context, workflowID string)
...
 		// Workflow is already in a terminal state, rollback and return
 		if err := tx.Rollback(ctx); err != nil {
 			return fmt.Errorf("failed to commit transaction: %w", err)
 		}
 		return nil
 	}
 
-	//Set the workflow's status to CANCELLED
+	// Set the workflow's status to CANCELLED and started_at_epoch_ms to NULL (so it does not block the queue, if any)
+	// XXX set dedup id to null? Fix it when we implement deduplication
```
maxdml: Will revisit this bit when I get to queue deduplication -- doesn't impact anything for now.

dbos/workflow.go
```diff
@@ -538,45 +526,62 @@ func (c *dbosContext) RunAsWorkflow(_ DBOSContext, fn WorkflowFunc, input any, o
...
+
+	// Compute the timeout based on the deadline
+	var timeout time.Duration
+	if !deadline.IsZero() {
+		timeout = time.Until(deadline)
+		/* unclear to me if this is a real use case:
+		if timeout < 0 {
+			return nil, newWorkflowExecutionError(workflowID, "deadline is in the past")
+		}
+		*/
```
maxdml, line 543: `timeout < 0` if the deadline is in the past. In this case the cancellation logic should already have been started. I might do something like checking the cancellation map if we hit that case -- we'll see.

dbos/workflow.go
```diff
@@ -607,37 +612,68 @@ func (c *dbosContext) RunAsWorkflow(_ DBOSContext, fn WorkflowFunc, input any, o
...
-		recordErr := c.systemDB.UpdateWorkflowOutcome(c.ctx, updateWorkflowOutcomeDBInput{
+
+		// If the afterFunc has started, the workflow was cancelled and the status should be set to cancelled
+		if stopFunc != nil && !stopFunc() {
+			// Wait for the cancel function to complete
+			// Note this must happen before we write on the outcome channel (and signal the handler's GetResult)
+			<-cancelFuncCompleted
+			// Set the status to cancelled and move on so we still record the outcome in the DB
+			status = WorkflowStatusCancelled
+		}
```
maxdml, line 660: For whoever is interested in how this stuff works: https://pkg.go.dev/context#AfterFunc

dbos/workflows_test.go
```diff
@@ -1961,3 +1962,376 @@ func TestSleep(t *testing.T) {
...
+		childHandle, err := RunAsWorkflow(ctx, waitForCancelWorkflow, "child-wait-for-cancel")
+		if err != nil {
+			t.Fatalf("failed to start child workflow: %v", err)
+		}
+
+		// Wait for the child workflow to complete
+		result, err := childHandle.GetResult()
+		if !errors.Is(err, context.DeadlineExceeded) {
+			t.Fatalf("expected child workflow to be cancelled, got: %v", err)
+		}
```
maxdml, line 2180: Note that the exception is whatever the child decided to return -- this is a direct handle that reads from the Child's outcome channel (not a polling handle that'd return `AwaitedWorkflowCancelled`)

dbos/workflows_test.go
```diff
@@ -1961,3 +1962,376 @@ func TestSleep(t *testing.T) {
...
+		status, err := handle.GetStatus()
+		if err != nil {
+			t.Fatalf("failed to get workflow status: %v", err)
+		}
+		if status.Status != WorkflowStatusCancelled {
+			t.Fatalf("expected workflow status to be WorkflowStatusCancelled, got %v", status.Status)
+		}
+	})
+
+	t.Run("RecoverWaitForCancelWorkflow", func(t *testing.T) {
```
maxdml, line 2285: I don't find this test super useful -- yes it checks that you can recover (run again) a workflow that has a timeout, and that this new invocation also times out, but it's exactly what you'd expect. At best this is another coverage of the durable timeout (the recovered workflow is started with the root context, that has no timeout)

PR #48, opened 2025-08-07
Fix child steps ids and get result recording

maxdml:
- Fix child step ID generation
- Fix test
- Rename some dangling names
- Fix recording of getResult as a step in a polling handle

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -638,19 +638,18 @@ func (c *dbosContext) RunAsWorkflow(_ DBOSContext, fn WorkflowFunc, input any, o
 		result, err := fn(workflowCtx, input)
 		status := WorkflowStatusSuccess
 
-		// If the afterFunc has started, the workflow was cancelled. Set the outcome
+		// If an error occurred, set the status to error
+		if err != nil {
+			status = WorkflowStatusError
+		}
+
+		// If the afterFunc has started, the workflow was cancelled and the status should be set to cancelled
 		if stopFunc != nil && !stopFunc() {
 			// Wait for the cancel function to complete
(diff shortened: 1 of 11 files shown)


dbos/workflow.go
```diff
@@ -154,17 +152,17 @@ func (h *workflowPollingHandle[R]) GetResult() (R, error) {
 			return *new(R), newWorkflowUnexpectedResultType(h.workflowID, fmt.Sprintf("%T", new(R)), fmt.Sprintf("%T", result))
 		}
 		// If we are calling GetResult inside a workflow, record the result as a step result
-		parentWorkflowState, ok := ctx.Value(workflowStateKey).(*workflowState)
-		isChildWorkflow := ok && parentWorkflowState != nil
-		if isChildWorkflow {
+		workflowState, ok := h.dbosContext.Value(workflowStateKey).(*workflowState)
```
maxdml, line 155: bug: we were using the wrong context and not recording `getResult` as a step.

dbos/workflow.go
```diff
@@ -491,11 +489,16 @@ func (c *dbosContext) RunAsWorkflow(_ DBOSContext, fn WorkflowFunc, input any, o
 	parentWorkflowState, ok := c.Value(workflowStateKey).(*workflowState)
 	isChildWorkflow := ok && parentWorkflowState != nil
 
+	if isChildWorkflow {
+		// Advance step ID if we are a child workflow
+		parentWorkflowState.NextStepID()
+	}
```
maxdml, line 495: Bug: `stepID` was incremented only if the workflow ID wasn't provided (so child workflows with custom wf ID wouldn't have the right step ID)

PR #49, opened 2025-08-07
Revamp steps ux

maxdml:
Here is a version of `RunAsStep` that accepts `any` functions. It still requires a single, typed return value, and an error.

The downside of this flexible UX is we need to perform reflection before calling the function, and to return the results. Reflection on this type of code is extremely hazardous, hence this PR:
- Minimize the use of reflection, does minimal checking
- Encapsulate all reflection code in a `safeExecute` function which recovers from panic
- Encapsulate the function call itself in `safeExecute` (something we should do anyways)

The PR tests:
- Valid signatures: e.g., no params, two params, custom struct return...
- Invalid inputs: not a function, nil function, direct nil input
- Invalid context (the first expected parameter of a step)
- Invalid returns (not R, error)
- Mismatching inputs vs provided function signature

Also add tests around recovering a step that panics -- and getting a proper error from the handle's `GetResult`.

Things we lose:
- Users must cast RunAsStep, because we can't infer the return type from the function signature anymore
- Users must register input structs for gob encoding. We might be able to do registration for them, but that's, again, lots of complex reflection code

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -327,6 +327,16 @@ func stepRetryWorkflow(dbosCtx DBOSContext, input string) (string, error) {
 	return RunAsStep[string](stepCtx, stepRetryAlwaysFailsStep, input)
 }
 
+// Helper struct and method for bound method test
+// XX do we need this declared here?
+type TestMethodStructForEdgeCase struct {
+	Value string
+}
+
+func (t TestMethodStructForEdgeCase) ProcessValue() (string, error) {
+	return "processed-" + t.Value, nil
+}
+
 func TestSteps(t *testing.T) {
 	dbosCtx := setupDBOS(t, true, true)
 
@@ -451,10 +461,8 @@ func TestSteps(t *testing.T) {
 		}
 	})
 
-	// Test comprehensive function signature validation for RunAsStep
+	// Test various happy/failure paths when calling RunAsStep with `any` functions
 	t.Run("RunAsStepSignatureValidation", func(t *testing.T) {
-		// Helper functions for testing different signatures
-
 		// Valid signatures
 		validNoParam := func(ctx context.Context) (string, error) { return "valid", nil }
 		validOneParam := func(ctx context.Context, input string) (string, error) { return input, nil }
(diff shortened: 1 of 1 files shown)


dbos/workflow.go
```diff
@@ -729,7 +737,8 @@ func setStepParamDefaults(params *StepParams, stepName string) *StepParams {
 
 var typeErasedStepNameToStepName = make(map[string]string)
 
-func RunAsStep[P any, R any](ctx DBOSContext, fn GenericStepFunc[P, R], input P) (R, error) {
+func RunAsStep[R any](ctx DBOSContext, fn any, input ...any) (R, error) {
```
kraftp, line 745: Does the function have to be of any type, or can we constrain its output but not its input?

dbos/workflow.go
```diff
@@ -738,37 +747,106 @@ func RunAsStep[P any, R any](ctx DBOSContext, fn GenericStepFunc[P, R], input P)
...
-	})
+	// Check if fn is a valid function
+	fnType := reflect.TypeOf(fn)
+	if fnType.Kind() != reflect.Func {
+		return *new(R), fmt.Errorf("fn must be a function, got %T", fn)
+	}
 
-	typeErasedStepNameToStepName[runtime.FuncForPC(reflect.ValueOf(typeErasedFn).Pointer()).Name()] = runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
+	// Check that the first argument implements context.Context
+	if fnType.NumIn() == 0 || fnType.In(0) != reflect.TypeOf((*context.Context)(nil)).Elem() {
```
kraftp: This is statically guaranteed, no?

dbos/workflow.go
maxdml: Right, we cannot constraint "just the output" and allow for a signature with unconstrained input. The closest to unconstrained is 

`func RunAsStep[R any](ctx DBOSContext, fn func(ctx context.Context, ...any)(R, error) , input ...any) (R, error)
`

Where the user would _have to_ provide a function signed `func(ctx context.Context, ...any)(R, error)`

dbos/workflow.go
maxdml: No the function is of type `any`, so we have no static guarantees about what the first argument here.

I can simplify these checks to wrap everything under the "safe execution" logic, which recovers from panics. Not exactly sure what the error message would be. (Tho I'd be inclined to simplify this code as much as possible.)

PR #50, opened 2025-08-08
accept a closure in RunAsStep

maxdml:
This version of a more flexible `RunAsStep` interface accepts a closure which only requires a context as first argument:

`type GenericStepFunc[R any] func(ctx context.Context) (R, error)`

User must pass the input type to `RunAsStep` (the compiler cannot infer the type from the closure)

Example usage:

```
stepCtx, stepCancelFunc := WithTimeout(ctx, 1*time.Millisecond)
defer stepCancelFunc() // Ensure we clean up the context
_, err := RunAsStep[string](stepCtx, func(context context.Context) (string, error) {
    return waitForCancelStep(context)
}) 
```

The part that's tricky is that we can't use the outer scope context in the step (here, the call to `waitForCancel`)

This is because we do some context augmentation inside DBOS. And we need to pass our new context to the function.

### DIFF
--- a/dbos/queues_test.go
+++ b/dbos/queues_test.go
@@ -30,14 +30,16 @@ This suite tests
 */
 
 func queueWorkflow(ctx DBOSContext, input string) (string, error) {
-	step1, err := RunAsStep(ctx, queueStep, input)
+	step1, err := RunAsStep[string](ctx, func(context context.Context) (string, error) {
+		return queueStep(context, input)
+	})
 	if err != nil {
 		return "", fmt.Errorf("failed to run step: %v", err)
 	}
 	return step1, nil
 }
 
-func queueStep(ctx context.Context, input string) (string, error) {
+func queueStep(_ context.Context, input string) (string, error) {
 	return input, nil
 }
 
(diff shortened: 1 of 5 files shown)


dbos/workflow.go
```diff
@@ -729,7 +729,7 @@ func setStepParamDefaults(params *StepParams, stepName string) *StepParams {
 
 var typeErasedStepNameToStepName = make(map[string]string)
 
-func RunAsStep[P any, R any](ctx DBOSContext, fn GenericStepFunc[P, R], input P) (R, error) {
+func RunAsStep[P any, R any](ctx DBOSContext, fn GenericStepFunc[R]) (R, error) {
```
kraftp: What is P? And why can't R be inferred from fn?

maxdml: Ops P is nothing :3

maxdml: That's why the type inference didn't work 😓

dbos/queues_test.go
```diff
@@ -30,14 +30,16 @@ This suite tests
 */
 
 func queueWorkflow(ctx DBOSContext, input string) (string, error) {
-	step1, err := RunAsStep(ctx, queueStep, input)
+	step1, err := RunAsStep[string](ctx, func(context context.Context) (string, error) {
```
qianl15: So do we still have to pass in `[string]` input type here? Or can it be inferred?

maxdml: nope this is because I forgot to remove one of the generic in the `RunAsStep` signature -- I clearly wasn't paying enough attention. Sorry about that.

qianl15 (approved):

PR #51, opened 2025-08-11
More tests

maxdml:


dbos/dbos.go
```diff
@@ -61,7 +61,7 @@ type DBOSContext interface {
 	Cancel()
 
 	// Workflow operations
-	RunAsStep(_ DBOSContext, fn StepFunc, input any) (any, error)
+	RunAsStep(_ DBOSContext, fn StepFunc) (any, error)
```
maxdml, line 64: Should have been in #50  😓

dbos/workflow.go
```diff
@@ -721,13 +735,22 @@ func setStepParamDefaults(params *StepParams, stepName string) *StepParams {
...
+var (
+	typeErasedStepNameToStepName = make(map[string]string)
+	typeErasedStepNameMutex      sync.RWMutex
+)
+
+func getTypeErasedStepName(stepName string) string {
+	typeErasedStepNameMutex.RLock()
+	defer typeErasedStepNameMutex.RUnlock()
+	return typeErasedStepNameToStepName[stepName]
+}
```
maxdml, line 753: Fix bug: concurrent access to this global map

dbos/workflows_test.go
```diff
@@ -142,11 +148,11 @@ func TestWorkflowsRegistration(t *testing.T) {
 				result, err := handle.GetResult()
 				_, err2 := handle.GetResult()
 				if err2 == nil {
-					t.Fatal("Second call to GetResult should return an error")
+					return nil, fmt.Errorf("Second call to GetResult should return an error")
```
maxdml, line 145: `t.Fatal` terminates the goroutine, but this is an inner goroutine ran by the test goroutine itself. The resulting error is quite hard to parse _and_ the test goroutine will move on to its cleanup actions, potentially deadlocking w/o much information. So, as a rule, we should only call `t.Fatal` in the "top level" test goroutine.

dbos/workflow.go
```diff
@@ -719,15 +738,17 @@ func setStepParamDefaults(params *StepParams, stepName string) *StepParams {
...
+		if value, ok := typeErasedStepNameToStepName.Load(stepName); ok {
+			params.StepName = value.(string)
+		}
 	}
 
 	return params
 }
 
-var typeErasedStepNameToStepName = make(map[string]string)
+var typeErasedStepNameToStepName sync.Map
```
maxdml, line 751: This map should be read-mostly. The first invocation of a step will store the reflection name.

dbos/workflow.go
```diff
@@ -204,10 +205,11 @@ type WrappedWorkflowFunc func(ctx DBOSContext, input any, opts ...WorkflowOption
 type workflowRegistryEntry struct {
 	wrappedFunction WrappedWorkflowFunc
 	maxRetries      int
+	name            string
 }
 
 // Register adds a workflow function to the registry (thread-safe, only once per name)
-func registerWorkflow(ctx DBOSContext, workflowName string, fn WrappedWorkflowFunc, maxRetries int) {
+func registerWorkflow(ctx DBOSContext, workflowFQN string, fn WrappedWorkflowFunc, maxRetries int, customName string) {
```
kraftp, line 212: What's going on here? What's the difference between FQN and customName?

kraftp: And the tests seem to use neither?

maxdml: The FQN is the reflection name, which we get with  `runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()`. 

The FQN is the only information we get at runtime when a user calls `RunAsWorkflow`.

A user can set a custom name when registering a workflow. There are now tests exercising this path.

The custom name is stored in the registry, alongside other registration-time parameters like `maxRetries`.

`RunAsWorkflow` resolve the workflow name at runtime when looking up registration options (from the registry.)

kraftp (approved):

PR #53, opened 2025-08-12
DBOS "client context"

maxdml:
This PR adds workflow management features usable by a "client context", that is, a DBOSContext object that is initialized but not launched. The benefits of doing so is being able to interact with DBOS workflows without having the process dequeue tasks, recover workflows, or run an admin server.

Features added:
- Resume workflow
- Fork workflow
- Enqueue workflow by name
- Add ability to dismiss output/input from list workflows

Because the registry is indexed by workflow FQN, we need to keep another map from workflow "custom" name to workflow FQN. This should have been tested in PR that added custom workflow names.

### DIFF
--- a/dbos/client_test.go
+++ b/dbos/client_test.go
@@ -632,3 +632,306 @@ func TestForkWorkflow(t *testing.T) {
 		t.Fatal("expected queue entries to be cleaned up after fork workflow tests")
 	}
 }
+
+func TestListWorkflows(t *testing.T) {
+	// Setup server context
+	serverCtx := setupDBOS(t, true, true)
+
+	// Create queue for communication
+	queue := NewWorkflowQueue(serverCtx, "list-workflows-queue")
+
+	// Simple test workflow
+	type testInput struct {
+		Value int
+		ID    string
+	}
+
+	simpleWorkflow := func(ctx DBOSContext, input testInput) (string, error) {
+		if input.Value < 0 {
+			return "", fmt.Errorf("negative value: %d", input.Value)
+		}
+		return fmt.Sprintf("result-%d-%s", input.Value, input.ID), nil
+	}
+	RegisterWorkflow(serverCtx, simpleWorkflow, WithWorkflowName("SimpleWorkflow"))
+
+	// Launch server
+	err := serverCtx.Launch()
+	if err != nil {
+		t.Fatalf("failed to launch server DBOS instance: %v", err)
+	}
+
+	// Setup client context
+	clientCtx := setupDBOS(t, false, false)
+
+	t.Run("ListWorkflowsFiltering", func(t *testing.T) {
+		var workflowIDs []string
+		var handles []WorkflowHandle[string]
+
+		// Record start time for filtering tests
+		testStartTime := time.Now()
+
+		// Start 10 workflows at 100ms intervals with different patterns
+		for i := range 10 {
+			var workflowID string
+			var handle WorkflowHandle[string]
+
+			if i < 5 {
+				// First 5 workflows: use prefix "test-batch-" and succeed
+				workflowID = fmt.Sprintf("test-batch-%d", i)
+				handle, err = Enqueue[testInput, string](clientCtx, GenericEnqueueOptions[testInput]{
+					WorkflowName:       "SimpleWorkflow",
+					QueueName:          queue.Name,
+					WorkflowID:         workflowID,
+					WorkflowInput:      testInput{Value: i, ID: fmt.Sprintf("success-%d", i)},
+					ApplicationVersion: serverCtx.GetApplicationVersion(),
+				})
+			} else {
+				// Last 5 workflows: use prefix "test-other-" and some will fail
+				workflowID = fmt.Sprintf("test-other-%d", i)
+				value := i
+				if i >= 8 {
+					value = -i // These will fail
+				}
+				handle, err = Enqueue[testInput, string](clientCtx, GenericEnqueueOptions[testInput]{
+					WorkflowName:       "SimpleWorkflow",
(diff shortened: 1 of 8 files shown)


dbos/system_database.go
```diff
@@ -297,6 +299,17 @@ func (s *sysDB) insertWorkflowStatus(ctx context.Context, input insertWorkflowSt
...
+	var applicationVersion *string
+	if len(input.status.ApplicationVersion) > 0 {
+		applicationVersion = &input.status.ApplicationVersion
+	}
+
+	var deduplicationID *string
+	if len(input.status.DeduplicationID) > 0 {
+		deduplicationID = &input.status.DeduplicationID
+	}
+
```
maxdml, line 312: I don't want to play with pointers in the codebase where we can distinguish variables through their zero values. Some rare cases, like the `offset` of listing workflows, must use pointers because `0` (the zero value of an int) is a valid offset.

dbos/system_database.go
```diff
@@ -671,6 +728,184 @@ func (s *sysDB) cancelWorkflow(ctx context.Context, workflowID string) error {
...
+						      workflow_deadline_epoch_ms = NULL, deduplication_id = NULL, 
+						      started_at_epoch_ms = NULL, updated_at = $4
+						  WHERE workflow_uuid = $5`
+
+	_, err = tx.Exec(ctx, updateStatusQuery,
+		WorkflowStatusEnqueued,
+		_DBOS_INTERNAL_QUEUE_NAME,
+		0,
+		time.Now().UnixMilli(),
+		workflowID)
```
maxdml: I am noticing that this resets the deadline (but keeps the timeout). In the current Go implementation, the timeout is not considered -- we only look at the deadline -- when running the workflow. I will fix this in a future PR.

dbos/system_database.go
```diff
@@ -671,6 +728,184 @@ func (s *sysDB) cancelWorkflow(ctx context.Context, workflowID string) error {
...
+	// Validate that startStep doesn't exceed the workflow's actual steps
+	maxStepQuery := `SELECT COALESCE(MAX(function_id), 0) FROM dbos.operation_outputs WHERE workflow_uuid = $1`
+	var maxStepID int
+	err = tx.QueryRow(ctx, maxStepQuery, input.originalWorkflowID).Scan(&maxStepID)
+	if err != nil {
+		return fmt.Errorf("failed to query max step ID: %w", err)
+	}
+	if input.startStep > maxStepID && maxStepID > 0 {
+		return fmt.Errorf("startStep %d exceeds workflow's maximum step %d", input.startStep, maxStepID)
+	}
```
maxdml: We might not need this check. I think it's useful for the user to know that their requested step doesn't exist. The Python behavior defaults to copying "steps up to N"

dbos/workflow.go
```diff
@@ -371,6 +383,7 @@ func RegisterWorkflow[P any, R any](ctx DBOSContext, fn GenericWorkflowFunc[P, R
 		// This type check is redundant with the one in the wrapper, but I'd better be safe than sorry
 		typedInput, ok := input.(P)
 		if !ok {
+			// FIXME: we need to record the error in the database here
```
maxdml, line 386: Will fix this in another PR: if the wrapped workflow is invoked from the registry, and given a wrong input type, it'd error silently. This case can happen if the client enqueues a workflow with the wrong input type.

dbos/workflow.go
```diff
@@ -1228,6 +1244,11 @@ func RetrieveWorkflow[R any](ctx DBOSContext, workflowID string) (workflowPollin
 	if ctx == nil {
 		return workflowPollingHandle[R]{}, errors.New("dbosCtx cannot be nil")
 	}
+
+	// Register the output for gob encoding
+	var r R
+	gob.Register(r)
```
maxdml, line 1304: We need to perform gob encoding here, because the client might not be running in the same program where the workflow was registered -- and automatic registration was performed -- and I eventually want to move the encoder into the `DBOSContext`.

dbos/system_database.go
kraftp: Resetting the deadline but keeping the timeout is the intended behavior of resume. The workflow should calculate a new deadline when it runs again (and if it doesn't, where are deadlines coming from?)

dbos/system_database.go
kraftp: We don't need this check, it's complex and doesn't add much

dbos/client_test.go
```diff
@@ -0,0 +1,946 @@
...
+		// Wait for workflow to signal it has started and step one completed
+		workflowStarted.Wait()
+
+		// Verify step one completed but step two hasn't
+		if stepsCompleted != 1 {
+			t.Fatalf("expected steps completed to be 1, got %d", stepsCompleted)
+		}
+
+		// Cancel the workflow
+		err = clientCtx.CancelWorkflow(workflowID)
```
kraftp: It will be confusing for some core operations to be methods on the context and others to be their own functions. I know why (generics), but it's still confusing. They should all have their own functions.

kraftp (approved):

dbos/client_test.go
maxdml: Agreed -- we should have a full 1:1 mapping.

dbos/system_database.go
maxdml: right now the deadlines are only coming from:
- The context if first run/enqueue
- The deadline field in the DB if the workflow already exists

Recomputing the deadline is only needed when the workflow is cancelled/resumed. In all other paths (e.g., recovery), the original deadline is the one to enforce.

I think the best way to do this will be to just recompute the deadline from the timeout in the resume operation.

PR #56, opened 2025-08-14
Bag o stuff

maxdml:
Major fix: deadline computation for enqueued workflows. Previously we'd actually set the deadline when the workflow was enqueued, not dequeued. This PR fixes it by considering the timeout when setting the final deadline. (Note: we also revert the change from #53 and let `resumeWorkflow` clear the deadline. It'll be recomputed from the timeout in `RunAsWorkflow`.)

Also:
- Add a test `TimeoutOnlySetOnDequeue`, to verify the above fix
- Record a workflow error when the function is called with the wrong parameter types. This could happen if the client enqueues a workflow with the wrong input type.
- Prevent the spawning of child workflows within steps and add a test (`ChildWorkflowCannotBeSpawnedFromStep`)
- Add a GHA to perform vulnerability checks and static analysis
- Set toolchain to 1.25 (and update test GHA accordingly). Note this is *not* the minimum version required by library consumers.
- Run `go vet` in test GHA
- Add "join discord" badge to README
- Make hardcoded values package level constants
- Fix a small bug where returning the admin server object would be missing the port
- Handle errors when writing the response in the healthcheck admin endpoint
- Set `ReadHeaderTimeout` on the admin server to prevent Slowloris attacks
- Handle sub millisecond timeouts (round to next millisecond)
- Handle negative timeouts (can happen with propagation delays): set timeout to 1ms
- Use a sync.Map for the workflow registry
- Check the executable is a regular file before opening it to compute its hash
- Remove unused WorkflowFunctionNotFound -- I don't see a use case for it
- Use `struct{}{}` instead of `bool` for notification channels (`struct{}{}` use zero space)
- Use an input struct for `dequeueWorkflows`
- Cleanup handles implementation with a base handler that implement once shared logic (`GetStatus`, `GetWorkflowID`).   other handles compose with this struct.
- Handle errors when failing to close notification listener connection
- Handle error when closing the connection used to set `LISTEN` channels in Postgres
- Handle errors (and print a log entry) when failing to run a scheduled workflow
- Fix possible integer overflows when converting uint to int (for queue priorities and fork start step)
- Complete DLQ test: resume the DLQ-ed workflow and check it can now run to completion. Check that the completed workflow can be ran many times past the retry limit.
- Pass `workflowCustomNametoFQN` to new contexts


dbos/workflow.go
```diff
@@ -177,7 +201,7 @@ func (h *workflowPollingHandle[R]) GetResult() (R, error) {
 			}
 			recordResultErr := h.dbosContext.(*dbosContext).systemDB.recordChildGetResult(h.dbosContext, recordGetResultInput)
 			if recordResultErr != nil {
-				// XXX do we want to fail this?
+				// Note: we might want GetResult to return this error if it happens, instead of returning the real error to the program.
```
maxdml: I'm seriously considering this -- we don't want the program to happily continue doing stuff if we weren't able to durably record `GetResult` as a step of a parent workflow.

kraftp: Sure, return the error here, it should be blocking as it indicates a serious database connection issue

kraftp (approved):
These changes and new tests look good. PR is large enough that I can't claim I caught everything, but high-level is good.

PR #58, opened 2025-08-15
Add a few missing endpoints to admin server

maxdml:
Still missing:
- garbage collection
- global timeout
- deactivate endpoint

Not sure we want the restart endpoint but I added it for backward compatibility

### DIFF
--- a/dbos/admin_server.go
+++ b/dbos/admin_server.go
@@ -24,7 +24,6 @@ const (
 	_WORKFLOW_STEPS_PATTERN           = "GET /workflows/{id}/steps"
 	_WORKFLOW_CANCEL_PATTERN          = "POST /workflows/{id}/cancel"
 	_WORKFLOW_RESUME_PATTERN          = "POST /workflows/{id}/resume"
-	_WORKFLOW_RESTART_PATTERN         = "POST /workflows/{id}/restart"
 	_WORKFLOW_FORK_PATTERN            = "POST /workflows/{id}/fork"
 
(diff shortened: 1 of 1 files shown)


dbos/admin_server_test.go
maxdml, line 1: The only tests I added are around list workflows:
- Check the timestamp validation (we expect UTC timestamps in the request that are converted to `time.Time`)
- Check that empty bodies are equated to no filters

dbos/admin_server.go
```diff
@@ -6,29 +6,166 @@ import (
...
+	_WORKFLOW_QUEUES_METADATA_PATTERN = "GET /dbos-workflow-queues-metadata"
+	_GARBAGE_COLLECT_PATTERN          = "POST /dbos-garbage-collect"
+	_GLOBAL_TIMEOUT_PATTERN           = "POST /dbos-global-timeout"
+	_QUEUED_WORKFLOWS_PATTERN         = "POST /queues"
+	_WORKFLOWS_PATTERN                = "POST /workflows"
+	_WORKFLOW_PATTERN                 = "GET /workflows/{id}"
+	_WORKFLOW_STEPS_PATTERN           = "GET /workflows/{id}/steps"
+	_WORKFLOW_CANCEL_PATTERN          = "POST /workflows/{id}/cancel"
+	_WORKFLOW_RESUME_PATTERN          = "POST /workflows/{id}/resume"
+	_WORKFLOW_RESTART_PATTERN         = "POST /workflows/{id}/restart"
```
kraftp: No restart endpoint, it's now deprecated everywhere

PR #59, opened 2025-08-16
Functional options for steps

maxdml:
Our current method for capturing the reflection name of a step is flawed. The typed erase function (anonymous) always has the same name, and thus every call to RunAsStep overwrites it. This of course results in wrong data in the `operation_results` table.

There is no good way to associated a unique key that the interface method can use on this map. Hence, let's just move to the cleaner functional options interface to handle step options, voiding the need for the in-memory map.

Added a test (`TestSteps/checkStepName`) that fails without this change.

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -75,7 +75,7 @@ type DBOSContext interface {
 	Cancel()       // Gracefully shutdown the DBOS runtime, waiting for workflows to complete and cleaning up resources
 
 	// Workflow operations
-	RunAsStep(_ DBOSContext, fn StepFunc) (any, error)                                                            // Execute a function as a durable step within a workflow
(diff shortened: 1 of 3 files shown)


kraftp (approved):
Good fix, but why did no existing test catch this bug?

kraftp (commented):
Also, could anything else have a similar problem? What about workflows?

maxdml:
> Good fix, but why did no existing test catch this bug?

Because we never had a tests that checked the step names after consecutive workflow / run a step invocations.

That being said I wasn't able to reproduce the exact bug the widget store showed. The problem became quickly obvious when debugging, so I moved on to the fix.

> could anything else have a similar problem? What about workflows?

Thank you for raising the question, there is actually an additional bug where a user could register two different workflows under the same custom name. Which I fixed in [6f8ced2](https://github.com/dbos-inc/dbos-transact-go/pull/59/commits/6f8ced2af74d3c46cea3d9bfd7a8726701e8f7e3).

PR #60, opened 2025-08-17
actually check DLQ

maxdml:
- One more pass at cleaning tests for testify: remove useless if statements
- Fix workflow DLQ test (actually test that the workflow ends up in the DLQ after max retries)

### DIFF
--- a/dbos/queues_test.go
+++ b/dbos/queues_test.go
@@ -213,11 +213,10 @@ func TestWorkflowQueues(t *testing.T) {
 
 		// Check that the workflow hits DLQ after re-running max retries
 		handles := make([]WorkflowHandle[any], 0, dlqMaxRetries+1)
-		for i := range dlqMaxRetries {
+		for _ = range dlqMaxRetries {
 			recoveryHandles, err := recoverPendingWorkflows(dbosCtx.(*dbosContext), []string{"local"})
 			require.NoError(t, err, "failed to recover pending workflows")
 			assert.Len(t, recoveryHandles, 1, "expected 1 handle")
-			t.Logf("Recovered handle: %v, index: %d", recoveryHandles[0].GetWorkflowID(), i)
 			dlqStartEvent.Wait()
 			dlqStartEvent.Clear()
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       run: go install gotest.tools/gotestsum@latest
 
     - name: Run tests
-      run: go vet ./... && gotestsum --format github-action -- -race ./...
+      run: go vet ./... && gotestsum --format github-action -- -race -v -count=1 ./...
```
maxdml, line 64: This disables "caching" for the tests...

dbos/queues_test.go
```diff
@@ -223,10 +224,25 @@ func TestWorkflowQueues(t *testing.T) {
...
+			require.NoError(t, err, "failed to get status of DLQ workflow handle")
+			if dlqStatus.Status != WorkflowStatusRetriesExceeded && retries < 10 {
+				time.Sleep(1 * time.Second) // Wait a bit before checking again
+				retries++
+				continue
 			}
+			require.NoError(t, err, "failed to get status of DLQ workflow handle")
+			assert.Equal(t, WorkflowStatusRetriesExceeded, dlqStatus.Status, "expected workflow to be in DLQ after max retries exceeded")
+			handles = append(handles, dlqHandle[0])
+			break
```
maxdml, line 244: we were not actually testing the DLQ behavior in this test -- now we do

dbos/workflow.go
```diff
@@ -572,7 +572,7 @@ func RunAsWorkflow[P any, R any](ctx DBOSContext, fn GenericWorkflowFunc[P, R],
 			}
 		}()
 
-		typedHandle := newWorkflowHandle[R](handle.dbosContext, handle.workflowID, typedOutcomeChan)
+		typedHandle := newWorkflowHandle(handle.dbosContext, handle.workflowID, typedOutcomeChan)
```
maxdml, line 575: not necessary because inferred from `typedOutcomeChan`

PR #62, opened 2025-08-19
flatten send and set event arguments

maxdml:
- Flatten send and set event inputs
- Simplify get /recv inputs
- Fix Sleep and Cancel interface signature to accept the context and match the package level methods ( 😢 )

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1116,7 +1116,11 @@ type RecvInput struct {
 	Timeout time.Duration // Maximum time to wait for a message
 }
 
-func (c *dbosContext) Recv(_ DBOSContext, input RecvInput) (any, error) {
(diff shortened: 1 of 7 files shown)


dbos/workflow.go
```diff
@@ -1167,49 +1158,40 @@ func Recv[R any](ctx DBOSContext, input WorkflowRecvInput) (R, error) {
...
-	return ctx.SetEvent(ctx, WorkflowSetEventInput{
-		Key:     input.Key,
-		Message: input.Message,
-	})
+	return ctx.SetEvent(ctx, key, message)
 }
 
-// WorkflowGetEventInput defines the parameters for retrieving an event from a workflow.
-type WorkflowGetEventInput struct {
+// GetEventInput defines the parameters for retrieving an event from a workflow.
```
kraftp: If we're going to make one in the pair flat, they should both be flat

kraftp (approved):

PR #71, opened 2025-08-20
Shutdown/Cancel interface

maxdml:


dbos/admin_server.go
```diff
@@ -541,18 +544,30 @@ func (as *adminServer) Start() error {
...
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		as.logger.Info("Admin server shutdown complete")
+	case <-ctx.Done():
+		as.logger.Warn("Admin server shutdown timed out")
+	}
+
```
maxdml, line 574: It turns out that `Shutdown` on `http.Server` is non blocking and graceful, i.e., it doesn't terminate existing connections. Waiting on the goroutine that started it to be done is a clean(er) way to wait for the server's full termination.

dbos/dbos.go
```diff
@@ -372,62 +373,90 @@ func (c *dbosContext) Launch() error {
...
+
+	// Wait for queue runner to finish
+	if c.queueRunner != nil {
+		c.logger.Info("Waiting for queue runner to complete")
+		select {
+		case <-c.queueRunner.completionChan:
+			c.logger.Info("Queue runner completed")
+			c.queueRunner = nil
+		case <-time.After(timeout):
+			c.logger.Warn("Timeout waiting for queue runner to complete", "timeout", timeout)
```
maxdml, line 418: we cannot safely set the pointer to `nil` in case of timeout: the resource might yet to be accessed, which would cause a panic.

dbos/dbos.go
```diff
@@ -372,62 +373,91 @@ func (c *dbosContext) Launch() error {
...
+// Cancel cancels the context, waits for workflows to complete, and stops the queue runner and workflow scheduler.
+// This is the core cancellation functionality that signals workflows should stop processing.
 // All workflows and steps contexts will be canceled, which one can check using their context's Done() method.
-//
-// This method blocks until all workflows finish and all resources are properly cleaned up.
-// It should be called when the application is shutting down to ensure data consistency.
-func (c *dbosContext) Cancel() {
-	c.logger.Info("Shutting down DBOS context")
+// Cancel is a permanent operation and should be called when the application is shutting down or deactivated.
+func (c *dbosContext) Cancel(timeout time.Duration) {
```
kraftp: This is very confusing. When would I as the end user want to use Cancel versus Shutdown? The only real use case is integration tests afaik, and then which one would I use?

dbos/admin_server.go
```diff
@@ -226,7 +227,7 @@ func newAdminServer(ctx *dbosContext, port int) *adminServer {
 	mux.HandleFunc(_DEACTIVATE_PATTERN, func(w http.ResponseWriter, r *http.Request) {
 		if as.isDeactivated.CompareAndSwap(0, 1) {
 			ctx.logger.Info("Deactivating DBOS executor", "executor_id", ctx.executorID, "app_version", ctx.applicationVersion)
-			// TODO: Stop queue runner, workflow scheduler, etc
+			ctx.Cancel(1 * time.Minute) // Cancel context, queue runner, and workflow scheduler
```
kraftp: This completely breaks queue draining on old versions

dbos/dbos.go
```diff
@@ -372,62 +373,91 @@ func (c *dbosContext) Launch() error {
...
 
 	// Cancel the context to signal all resources to stop
-	c.ctxCancelFunc(errors.New("DBOS shutdown initiated"))
+	c.ctxCancelFunc(errors.New("DBOS cancellation initiated"))
 
 	// Wait for all workflows to finish
 	c.logger.Info("Waiting for all workflows to finish")
-	c.workflowsWg.Wait()
-	c.logger.Info("All workflows completed")
+	done := make(chan struct{})
```
kraftp, line 397: Why do we wait for all workflows to finish? If we're doing that, what's the practical point of this (unless it's just for testing). You're usually using DBOS because it lets you write long-running workflows that can persist across restarts and such, but this just sits there and blocks (potentially forever, as you're still dequeueing new workflows while waiting for old ones to finish).

dbos/admin_server.go
maxdml: Good point, Cancel shouldn't stop the queue runner, just the workflow scheduler.

dbos/dbos.go
maxdml: For us there is a direct use case: deactivate. We don't want to destroy everything (e.g. the system db pool) during deactivate.

Generally speaking, I like the idea that users can get a very go-like cancel propagation to all their workflows/steps on a given process. `Cancel()` works very well for that.

dbos/dbos.go
maxdml: It is not mandatory (there is a timeout) but very useful for testing

dbos/dbos.go
maxdml: In hindsight, a better design would be to have Cancel only cancel the context, then optionally wait on the workflows.

Shutdown would shutdown everything

The deactivate endpoint can just repeat the logic to stop the scheduler

kraftp: I'm still confused, and like we said, Cancel doesn't do what deactivate wants (and if it's just for deactivate, it shouldn't be in the public interface).

dbos/dbos.go
kraftp: Got it, for testing and with a timeout this makes a lot of sense

dbos/dbos.go
maxdml: What I am suggesting is that:
- Deactivate just does its own thing: stop the workflow scheduler
- Cancel() just cancels the underlying context.Context
- Shutdown actually shuts down resources

maxdml: fixed in https://github.com/dbos-inc/dbos-transact-golang/pull/71/commits/db5f2248b414c7b5c6d85b15d1dffddd3f9f3a92

kraftp: I still don't see the use case for Cancel

maxdml: The use case is: assume you want to cancel all the running workflows in an application.

Cancelling a workflow using the DBOS cancel API results in a cancellation only when the workflow does a step. If the workflow is doing some longer running operation, it will never know.

This is a problem we have in all languages. But in go, we can exploit the native semantics of context for signaling cancellation, and expect that programmers will check on the state of the workflow/step input context in their code.

kraftp: I don't know, is this something we need to support with a top-level API? I've never seen anyone ask for it any any other language.

maxdml: I slept over this and I think the current design isn't what we need. Specifically, we have a single root DBOSContext that is used for the queue runner, system db, and the workflows. So cancelling it cancels all, not just the workflows.

I do think that having a way to signal cancellation to all the workflows would be a good, idiomatic way to signal that workflows should stop. That would require having more dedicated contexts. Not a priority!

maxdml: One thing is that I'm renaming `Cancel` for `Shutdown`, because `Cancel` will be confusing.

kraftp (approved):

PR #72, opened 2025-08-21
garbage collection, global timeout

maxdml:
also:
- set max/min conns and timeouts on sysdb pool
- fix a small race in a test, remove unused code


dbos/system_database.go
```diff
@@ -756,6 +774,87 @@ func (s *sysDB) cancelWorkflow(ctx context.Context, workflowID string) error {
...
+	}
+
+	return nil
+}
+
+func (s *sysDB) garbageCollectWorkflows(ctx context.Context, input garbageCollectWorkflowsInput) error {
+	// Validate input parameters
+	if input.rowsThreshold != nil && *input.rowsThreshold <= 0 {
+		return fmt.Errorf("rowsThreshold must be greater than 0, got %d", *input.rowsThreshold)
+	}
```
maxdml, line 810: Technically this could be enforced by using a unsigned integer, but I trust this makes for a better error message.

dbos/workflows_test.go
```diff
@@ -844,23 +844,24 @@ func TestChildWorkflow(t *testing.T) {
...
-
-			// Signal the child workflow is started
-			pollingHandleStartEvent.Set()
-
-			result, err := childHandle.GetResult()
-			if err != nil {
-				return "", fmt.Errorf("failed to get result from child workflow: %w", err)
-			}
-			return result, nil
+			return "", nil
```
maxdml, line 862: There is an interesting race here.

This test calls the workflow twice, then eventually waits on the workflow result before completing. After completion, tests call a cleanup function that calls `Shutdown`.

It is possible that the 2nd invocation of the workflow (typically the one where the child gets a polling handle) completes slow enough that the test is already in the cleanup phase.

This will result in a segfault, given that Shutdown clears the system DB.

Because this test just wants to verify the 2nd invocation of the child gets a polling handle, we can simply do the `GetResult() `on the first call only.

dbos/workflows_test.go
```diff
@@ -950,26 +951,6 @@ func idempotencyWorkflow(dbosCtx DBOSContext, input string) (string, error) {
...
-func idempotencyWorkflowWithStep(dbosCtx DBOSContext, input string) (int64, error) {
-	RunAsStep(dbosCtx, func(ctx context.Context) (int64, error) {
-		return incrementCounter(ctx, int64(1))
-	})
-	idempotencyWorkflowWithStepEvent.Set()
-	RunAsStep(dbosCtx, func(ctx context.Context) (string, error) {
-		return blockingStep(ctx)
-	})
-	return idempotencyCounter, nil
-}
```
maxdml, line 971: unused

PR #73, opened 2025-08-22
Conductor

maxdml:
Implement Conductor client.

# Design

A conductor object is attached to a `DBOSContext`. Its lifecycle management is done like other resources, inside `DBOSContext` `Launch` and `Shutdown` methods.

The conductor operates through two goroutines a "main" goroutine handling conductor messages, and a "ping" goroutine to keep the connection alive.

# Main goroutine

- Handles connecting to conductor
- Handles responding to conductor message
- Will stop only if its context is cancelled (root DBOSContext)

The main goroutine is entirely responsible for managing the connection: whenever it detects it should connect to the server, it'll start a ping goroutine. Whenever it gets an error (e.g., reading on the websocket or sending a message), it will set the connection for closing.

# Ping goroutine

- Periodically sends keepalive pings to the server
- Handles pong answers

The ping goutine will exit and signal the connection should be close if:
- Its context is cancelled or
- It gets an error trying to send the ping

# Others

- There cannot be concurrent writers on the websocket, so all write operations are protected by a mutex
- Shutdown logic: we cancel the ping goroutine, close the connection and wait for both goroutine to terminate.
- Add a missing executors ID filter on `ListWorkflows`

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -319,10 +319,7 @@ func NewDBOSContext(inputConfig Config) (DBOSContext, error) {
 
 	// Initialize conductor if API key is provided
 	if config.ConductorAPIKey != "" {
-		if initExecutor.executorID == "local" {
-			initExecutor.executorID = uuid.NewString()
-		}
-
+		initExecutor.executorID = uuid.NewString()
 		if config.ConductorURL == "" {
 			dbosDomain := os.Getenv("DBOS_DOMAIN")
 			if dbosDomain == "" {


dbos/conductor.go
```diff
@@ -0,0 +1,1007 @@
...
+	var filteredWorkflows []WorkflowStatus
+	if req.Body.QueueName == nil {
+		for _, wf := range workflows {
+			if wf.QueueName != "" {
+				filteredWorkflows = append(filteredWorkflows, wf)
+			}
+		}
+	} else {
+		filteredWorkflows = workflows
+	}
```
maxdml, line 704: Note that I'll eventually add a filter that'll go down the SQL query (while maintaining the simplified listing UX)

dbos/dbos.go
```diff
@@ -312,9 +314,35 @@ func NewDBOSContext(inputConfig Config) (DBOSContext, error) {
 	initExecutor.logger.Info("System database initialized")
 
 	// Initialize the queue runner and register DBOS internal queue
-	initExecutor.queueRunner = newQueueRunner()
+	initExecutor.queueRunner = newQueueRunner(initExecutor.logger)
 	NewWorkflowQueue(initExecutor, _DBOS_INTERNAL_QUEUE_NAME)
 
+	// Initialize conductor if API key is provided
+	if config.ConductorAPIKey != "" {
+		if initExecutor.executorID == "local" {
```
kraftp: It's safer to just always generate a new ID here. Otherwise, there's potential for trouble--multiple servers sharing the same ID, for example.

dbos/dbos.go
```diff
@@ -47,10 +51,12 @@ func processConfig(inputConfig *Config) (*Config, error) {
 	}
 
 	dbosConfig := &Config{
-		DatabaseURL: inputConfig.DatabaseURL,
-		AppName:     inputConfig.AppName,
-		Logger:      inputConfig.Logger,
-		AdminServer: inputConfig.AdminServer,
+		DatabaseURL:     inputConfig.DatabaseURL,
```
kraftp, line 54: Application version should also be an option here

maxdml: I'm adding this and executor ID in a different PR

PR #74, opened 2025-08-27
Executor config: app version/ executor ID

maxdml:
- allow user provided executor ID, app version
- make update wf outcome idempotent if status transitions from cancel to error


dbos/system_database.go
```diff
@@ -1646,7 +1646,7 @@ func (s *sysDB) recv(ctx context.Context, input recvInput) (any, error) {
 	}
 	recordedResult, err := s.checkOperationExecution(ctx, checkInput)
 	if err != nil {
-		return nil, fmt.Errorf("failed to check operation execution: %w", err)
+		return nil, err
```
maxdml, line 1649: This seemingly innocuous change will be dug deeper in a future PR: the end-user should receive, when appropriate, a `DBOSError`, rather than a normal error wrapping the `DBOSError`. Will need polish.

kraftp (approved):

PR #76, opened 2025-08-27
Add release script and GHA

maxdml:
- Add a script that creates a tag, bump the minor version, push the tag and creates/push a release branch.
- Major versions upgrade are not automatic -- but will work if we provide the new version as an input to the GHA
- Patches are not covered
- There is a mandatory gating to the action (YES!!!)

I tested manually the "clean and up to date local main" requirement, the YES!!! gating, minor version bumps (including from a valid semver `-alpha`, major version bumps)


dbos/admin_server_test.go
```diff
@@ -592,7 +592,7 @@ func TestAdminServer(t *testing.T) {
 
 			// Verify no new executions occurred
 			finalCount := executionCount.Load()
-			assert.Equal(t, countAfterDeactivate, finalCount,
+			assert.LessOrEqual(t, finalCount, countAfterDeactivate+1,
```
maxdml, line 595: Give some leeway to this test

kraftp (commented):
Is there any notion of preview releases in Go?

maxdml:
> Is there any notion of preview releases in Go?

No, releases, and specifically "latest", are based on (semver) tags. semver supports `-alpha`, `-beta` and so on.

You can also `go get` a specific commit and/or branch.

In our tests we do `go get @main` to get what we call "preview" in TS/Python

maxdml:
> This looks reasonable--the same as our Python versioning script as far as I can tell.

One difference is that we reset the patch version when bumping a minor version -- I don't think the python version does that?

PR #78, opened 2025-08-28
queue filter in sql

maxdml:
Add a new option to ListWorkflows to filter for "queues only".

### DIFF
--- a/dbos/admin_server.go
+++ b/dbos/admin_server.go
@@ -378,25 +378,16 @@ func newAdminServer(ctx *dbosContext, port int) *adminServer {
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


dbos/admin_server.go
```diff
@@ -378,27 +378,18 @@ func newAdminServer(ctx *dbosContext, port int) *adminServer {
 			}
 		}
 
-		req.Status = "" // We are not expecting a filter here but clear just in case
```
maxdml, line 381: Because the console never passes a status, I previously just dismissed any status filter received on the queues endpoint. I think this is wrong, so this PR only sets enqueued/pending if no status is present in the request.

PR #98, opened 2025-08-30
public GetWorkflowSteps

maxdml:
- Expose a public `GetWorkflowSteps` function
- Move internal queue creation at launch of the context
- Skip the dequeue query if `maxTasks` <= 0
- Make db migrations idempotent


dbos/migrations/000001_initial_dbos_schema.up.sql
```diff
@@ -7,7 +7,7 @@ CREATE SCHEMA IF NOT EXISTS dbos;
 CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 
 -- Create workflow_status table
-CREATE TABLE dbos.workflow_status (
+CREATE TABLE IF NOT EXISTS dbos.workflow_status (
```
kraftp, line 10: We should replace this entirely with the new migration system. Doesn't have to be in this PR, but should be done before release.

PR #99, opened 2025-08-30
Admin server port

shirzady1934:
Replaced hard-coded port `3001` with a configurable `AdminServerPort` field in `Config`.  
Updated `dbos.go` to use the provided port (defaulting to 3001) and adjusted tests to use a custom port (`38081`).  

✅ All `TestAdminServer` cases now pass.

### DIFF
--- a/dbos/admin_server_test.go
+++ b/dbos/admin_server_test.go
@@ -15,6 +15,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testAdminPort = 38081
+
 func TestAdminServer(t *testing.T) {
 	databaseURL := getDatabaseURL()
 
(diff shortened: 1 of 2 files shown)


dbos/admin_server_test.go
```diff
@@ -36,7 +38,7 @@ func TestAdminServer(t *testing.T) {
 
 		// Verify admin server is not running
 		client := &http.Client{Timeout: 1 * time.Second}
-		_, err = client.Get(fmt.Sprintf("http://localhost:3001/%s", strings.TrimPrefix(_HEALTHCHECK_PATTERN, "GET /")))
+		_, err = client.Get(fmt.Sprintf("http://localhost:%d/%s", testAdminPort, strings.TrimPrefix(_HEALTHCHECK_PATTERN, "GET /")))
```
af-md: code looks great!

I was wondering if it might be better to reuse the _DEFAULT_ADMIN_SERVER_PORT instead of _testAdminPort_  to follow the same pattern in the test suite of reusing constants. What do you think?

shirzady1934: I changed testAdminPort to _DEFAULT_ADMIN_SERVER_PORT.

PR #101, opened 2025-08-30
init exp backoff + jitter for system_database.go

### LINKED ISSUE #95: exponential backoff + jitter for notification listener
Currently we retry at a fixed interval (`_DB_CONNECTION_RETRY_DELAY`)

pranshu-raj-211:
Fixes #95 

Added exponential backoff delay + jitter to replace fixed delay being used in `shutdown` and `notificationListenerLoop`.

Replaced delay interval constant by constants

```go
    _DB_CONNECTION_RETRY_BASE_DELAY = 500 * time.Millisecond
    _DB_CONNECTION_RETRY_FACTOR = 2
    _DB_CONNECTION_RETRY_MAX_RETRIES = 3
    _DB_CONNECTION_MAX_DELAY = 10 * time.Second
```

### How this works
Func `backoffWithJitter` takes an argument attempt - attempts of backoff delay done yet.
Calculates the delay as 

`baseDelay * math.Pow(retryFactor, attempts) + jitter`

Jitter being calculated as a random value between 0.3 and 1.3 (constants of no particular significance, can change - repo has varying implementations of backoff with jitter).

If the delay calculate exceeds a cap defined in constants - take the constant instead.

If the number of retries get too large -> return an error saying so (optional, maybe want to timeout if takes too long).

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"math/rand"
 	"net/url"
-	"rand"
 	"strings"
 	"sync"
 	"time"
@@ -136,11 +136,11 @@ const (
 	_DBOS_WORKFLOW_EVENTS_CHANNEL = "dbos_workflow_events_channel"
 
 	// Database retry timeouts
-	_DB_CONNECTION_RETRY_BASE_DELAY = 500 * time.Millisecond
-	_DB_CONNECTION_RETRY_FACTOR = 2
+	_DB_CONNECTION_RETRY_BASE_DELAY  = 500 * time.Millisecond
+	_DB_CONNECTION_RETRY_FACTOR      = 2
 	_DB_CONNECTION_RETRY_MAX_RETRIES = 3
-	_DB_CONNECTION_MAX_DELAY = 10 * time.Second
-	_DB_RETRY_INTERVAL         = 1 * time.Second
+	_DB_CONNECTION_MAX_DELAY         = 10000 * time.Millisecond
+	_DB_RETRY_INTERVAL               = 1 * time.Second
 )
 
 func runMigrations(databaseURL string) error {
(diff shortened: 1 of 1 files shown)


dbos/system_database.go
```diff
@@ -2317,3 +2323,19 @@ func (qb *queryBuilder) addWhereLessEqual(column string, value any) {
...
+	if retryAttempt >= _DB_CONNECTION_RETRY_MAX_RETRIES {
+		return _DB_CONNECTION_RETRY_BASE_DELAY
+	}
+	exp := float64(_DB_CONNECTION_RETRY_BASE_DELAY) * math.Pow(_DB_CONNECTION_RETRY_FACTOR, float64(retryAttempt))
+	// cap delay to maximum of _DB_CONNECTION_MAX_DELAY milliseconds
+	if exp > float64(_DB_CONNECTION_MAX_DELAY) {
+		exp = float64(_DB_CONNECTION_MAX_DELAY)
+	}
+
+	jitter := 0.5 + rand.Float64()
```
af-md: Great job! 

However, jitter implementation may not provide intended randomisation

The current jitter adds a fixed 0.5-1.5ms to retry delays. Imo that doesn't provide enough range to retry again. Ideally the range should spread meaningfully between -25% and +25% of the exp. For example instead of 500.5-501.5ms for the first retry, we should aim between 375-625ms. A decent spread.

What do you think?

af-md (dismissed):

dbos/system_database.go
pranshu-raj-211: That makes sense, I was thinking of it in seconds, wrong unit.
Simple fix, pushing it now.

Where should I add tests for this, if it's required (built a basic table driven test)?

pranshu-raj-211: Also, I've made the delay be fixed (equal to base delay - 500ms) if the number of retries exceeds what we've set as the max retries, is this behavior desirable or should it be changed?

Issue #61, opened 2025-08-19
notificationListenerLoop backoffs


maxdml:
notificationListenerLoop should backoff when getting errors while waiting from PG NOTIFY

maxdml:
close as duplicate of #95

PR #103, opened 2025-09-01
Chore: add init exponential backoff + jitter for system_database.go

pranshu-raj-211:
Refer to PR #101  (got closed while renaming branch)

## What this PR does
Adds exponential backoff and jitter to replace fixed delays in system_database.go

## How it is implemented
Replaced delay interval constant by constants

```
    _DB_CONNECTION_RETRY_BASE_DELAY = 500 * time.Millisecond
    _DB_CONNECTION_RETRY_FACTOR = 2
    _DB_CONNECTION_RETRY_MAX_RETRIES = 3
    _DB_CONNECTION_MAX_DELAY = 10 * time.Second
```

Func backoffWithJitter takes an argument attempt - attempts of backoff delay done yet.
Calculates the delay as

```
exp = baseDelay * math.Pow(retryFactor, attempts)
jitter = 0.75 + rand.Float64()*0.5  // +-25% of exp to be jitter
delay = exp * jitter
```

Capping provided against delay intervals getting too large.

If number of retries exceeds a certain limit - fallback to fixed base delay (500ms)

Jitter added separately even for fixed intervals.

Constants may need to be changed.

## Tests
I do have a basic table driven test for this, unsure of which file to add it in.


dbos/system_database.go
```diff
@@ -2317,3 +2323,20 @@ func (qb *queryBuilder) addWhereLessEqual(column string, value any) {
 	qb.whereClauses = append(qb.whereClauses, fmt.Sprintf("%s <= $%d", column, qb.argCounter))
 	qb.args = append(qb.args, value)
 }
+
+func backoffWithJitter(retryAttempt int) time.Duration {
```
af-md, line 2381: You can test it by creating `system_database_test.go` file because the function is tightly coupled with the database retry properties.

I am also wondering if we should make a generic `backoffWithJitter` that could be re-used across the sdk?

maxdml: Re-using this function is a good idea, we do use it in other places already. But we can make it another PR.

dbos/system_database.go
```diff
@@ -134,8 +136,11 @@ const (
 	_DBOS_WORKFLOW_EVENTS_CHANNEL = "dbos_workflow_events_channel"
 
 	// Database retry timeouts
-	_DB_CONNECTION_RETRY_DELAY = 500 * time.Millisecond
-	_DB_RETRY_INTERVAL         = 1 * time.Second
+	_DB_CONNECTION_RETRY_BASE_DELAY  = 500 * time.Millisecond
+	_DB_CONNECTION_RETRY_FACTOR      = 2
+	_DB_CONNECTION_RETRY_MAX_RETRIES = 3
+	_DB_CONNECTION_MAX_DELAY         = 10000 * time.Millisecond
+	_DB_RETRY_INTERVAL               = 1 * time.Second
```
maxdml: We can be a bit more lenient on these:
Max delay: 2 minutes
Max retries: 10
Also the base retry delay should be = to the base interval

dbos/system_database.go
```diff
@@ -2317,3 +2323,20 @@ func (qb *queryBuilder) addWhereLessEqual(column string, value any) {
...
+	exp := float64(_DB_CONNECTION_RETRY_BASE_DELAY) * math.Pow(_DB_CONNECTION_RETRY_FACTOR, float64(retryAttempt))
+	// cap delay to maximum of _DB_CONNECTION_MAX_DELAY milliseconds
+	if exp > float64(_DB_CONNECTION_MAX_DELAY) {
+		exp = float64(_DB_CONNECTION_MAX_DELAY)
+	}
+
+	// want randomization between +-25% of exp
+	jitter := 0.75 + rand.Float64()*0.5
+	return time.Duration(exp * jitter)
+}
```
maxdml: Thanks this is a good start! We also need a way to scale down the interval after successful iterations (see an example in https://github.com/dbos-inc/dbos-transact-golang/blob/main/dbos/queue.go).

Generally speaking I am increasingly considering generalizing the retry utility, as @af-md suggested, potentially offloading it to https://github.com/avast/retry-go

dbos/system_database.go
pranshu-raj-211: By the base interval do you mean the `_DB_RETRY_INTERVAL`?
The issue #95 showed the variable that was being used for this, which in the code was `_DB_CONNECTION_RETRY_DELAY = 500 * time.Millisecond`, which I took as default.

Changed the other values, updating tests to match the same.

dbos/system_database.go
pranshu-raj-211: That makes sense.

Current design takes in an int retryAttempt, which offloads the responsibility of this to the caller, which keeps track of retries and is supposed to reset it when done.

As an example from notificationListenerLoop

```
retryAttempt := 0
	for {
		// Block until a notification is received. OnNotification will be called when a notification is received.
		// WaitForNotification handles context cancellation: https://github.com/jackc/pgx/blob/15bca4a4e14e0049777c1245dba4c16300fe4fd0/pgconn/pgconn.go#L1050
		err := s.notificationListenerConnection.WaitForNotification(ctx)
		if err != nil {
			// Context cancellation
			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
				s.logger.Info("Notification listener loop exiting due to context cancellation", "cause", context.Cause(ctx), "error", err)
				return
			}

			// Connection closed (during shutdown) - exit gracefully
			if s.notificationListenerConnection.IsClosed() {
				s.logger.Info("Notification listener loop exiting due to connection closure")
				return
			}

			// Other errors - log and retry.
			s.logger.Error("Error waiting for notification", "error", err)
			time.Sleep(backoffWithJitter(retryAttempt))
			retryAttempt += 1
			continue
		}
	}
```

How should I proceed with this?

dbos/system_database.go
maxdml: Let's make `_DB_CONNECTION_RETRY_BASE_DELAY` equal 1 second. Thanks!

dbos/system_database.go
```diff
@@ -288,7 +293,7 @@ func (s *sysDB) shutdown(ctx context.Context, timeout time.Duration) {
 	// Allow pgx health checks to complete
 	// https://github.com/jackc/pgx/blob/15bca4a4e14e0049777c1245dba4c16300fe4fd0/pgxpool/pool.go#L417
 	// These trigger go-leak alerts
-	time.Sleep(_DB_CONNECTION_RETRY_DELAY)
+	time.Sleep(backoffWithJitter(0))
```
maxdml: We can actually hard code this sleep to be 500 ms. It is totally unrelated to DB retries. (We just give some grace period to `pgx` to terminate.)

dbos/system_database.go
maxdml: For this PR let's just do the simple thing: if `err` is `nil` after `WaitForNotification`, and `retryAttempt > 0`, let's just decrement `retryAttempt`. This will do multiplicative decrease multiplicative increase (which is OK for this specific case.)

pranshu-raj-211:
Added the comment to ignore G404 - rand.
Also fixed issue causing test failure - used time.Second instead of time.Millisecond in the shutdown function, which caused a timeout.

pranshu-raj-211:
Resolved merge conflicts (resulting from #106 push to main).

Tested security and test actions locally (act, using medium size image config).

pranshu-raj-211:
Also, I was curious, how does this relate to #61

maxdml (approved):

maxdml:
> Also, I was curious, how does this relate to #61

its a duplicate.

Thanks for your contribution!

PR #105, opened 2025-09-02
draft CLI

maxdml:


cmd/dbos/workflow.go
```diff
@@ -0,0 +1,411 @@
...
+		return fmt.Errorf("failed to get workflow status: %w", err)
+	}
+
+	// Output results
+	if jsonOutput {
+		return outputJSON(status)
+	}
+
+	// Pretty print
+	fmt.Printf("Workflow ID: %s\n", status.ID)
```
kraftp: In the interest of maintainability (having learned this lesson in the other CLIs) I would get rid of all the non-JSON output formatting and only emit JSON.

cmd/dbos/reset.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to connect to postgres database: %w", err)
+	}
+	defer db.Close()
+
+	// Drop the system database if it exists
+	logger.Info("Resetting system database", "database", dbName)
+
+	// First, terminate all connections to the database
+	terminateQuery := fmt.Sprintf(`
```
kraftp: Replace this (in all places it's used) with DROP DATABASE WITH (FORCE).

kraftp (approved):
Looks good overall. We can fix serialization in another PR.

PR #106, opened 2025-09-04
dbos migration

maxdml:


dbos/system_database.go
```diff
@@ -139,52 +139,137 @@ const (
...
 
 	return nil
 }
 
+type migrationFile struct {
+	version int64
+	sql     string
+}
+
+func parseMigrationFiles() ([]migrationFile, error) {
```
kraftp: Let's not do any of this file parsing stuff. How does this even work in a compiled binary? Just make them an array in memory.

kraftp: Literally, this should be identical to the Python implementation (without the backwards-compatibility Alembic nonsense)

maxdml: What this code is doing is embedding the files at compile time in memory using the [embed](https://pkg.go.dev/embed) package.

kraftp: Let's just do strings. Way simpler.

maxdml: I don't think its simpler. sql files are much more manageable. We can embed them as strings, one by one, in the file if we need, so we don't have to do the sorting/"parsing", but having them in the files is really bad imo. We have to do it in the other languages because they are not compiled, but here we have the facility to do this both correctly and manageably

kraftp: I'm looking at this code and I think strings are simpler. You don't have to worry about parsing file names or handle errors if the migrations are just an array of strings. You can literally delete this entire function.

dbos/migrations/000001_initial_dbos_schema.down.sql
```diff
@@ -1,18 +0,0 @@
--- 001_initial_dbos_schema.down.sql
-
--- Drop triggers first
-DROP TRIGGER IF EXISTS dbos_notifications_trigger ON dbos.notifications;
-DROP TRIGGER IF EXISTS dbos_workflow_events_trigger ON dbos.workflow_events;
-
```
kraftp, line 6: Also, make sure the migration SQL is IDENTICAL to the one in Python. We want an identical system database across all four languages.

dbos/system_database.go
maxdml: fixed in https://github.com/dbos-inc/dbos-transact-golang/pull/106/commits/877acd0c1517712764ce3d42d58161f1e93b2af0

dbos/migrations/000001_initial_dbos_schema.down.sql
maxdml: fixed in https://github.com/dbos-inc/dbos-transact-golang/pull/106/commits/3a6aee06c5872561f8c9d845a657733815422831

dbos/system_database.go
kraftp: Ah that's very nice, best of both worlds

kraftp (approved):

PR #108, opened 2025-09-04
Remove temp install

maxdml:
- The CLI test can now use the latest golang SDK (had to specifically get a version with the new migration scheme before)
- Update a log info telling users to use `postgresql` instead of `postgres` as the db driver

### DIFF
--- a/cmd/dbos/postgres.go
+++ b/cmd/dbos/postgres.go
@@ -174,7 +174,7 @@ func startDockerPostgres() error {
 		return err
 	}
 
-	logger.Info("Postgres available", "url", fmt.Sprintf("postgresql://postgres:%s@localhost:5432", password))
+	logger.Info("Postgres available", "url", fmt.Sprintf("postgres://postgres:%s@localhost:5432", password))
 	return nil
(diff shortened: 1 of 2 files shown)


cmd/dbos/postgres.go
```diff
@@ -174,7 +174,7 @@ func startDockerPostgres() error {
 		return err
 	}
 
-	logger.Info("Postgres available", "url", fmt.Sprintf("postgresql://postgres:%s@localhost:5432", password))
+	logger.Info("Postgres available", "url", fmt.Sprintf("postgres://postgres:%s@localhost:5432", password))
```
qianl15, line 177: Why change this? I thought you want users to use `postgresql` not `postgres`

maxdml: In most go library, `postgres` and `postgresql` are interchangeable. But with some, only `postgres` is valid. I've had a connection error a couple times because of a wrong driver name. To be honest I don't know which exact library we use(d) triggered the error, but I think this will be helpful for people importing these libraries.

PR #109, opened 2025-09-04
Run as step inside go routines

af-md:
## Summary
Adds support for running steps inside goroutines with deterministic step ID generation.

This is by no means the final solution, it's a PR to get feedback.

## Problem
Currently, running steps inside goroutines causes non-deterministic step ID generation due to race conditions:

## Solution

Pre-generate step IDs before launching goroutines:

- Added Go() function that pre-generates step IDs and runs steps in goroutines
- Added WithNextStepID option to pass pre-generated IDs to RunAsStep
- Modified RunAsStep to use pre-generated IDs when provided

## Open Questions

- Does this satisfy the UX you had in mind @maxdml ? I am a bit unsure as to what you had in mind but gave it a go anyway. 
- Might be worth giving me fuller example of how you think the UX would work inside the workflow

## ToDos
- I still need to add some docs
- Still need to write tests for it as well

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -8,7 +8,6 @@ import (
 	"math"
 	"reflect"
 	"runtime"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -1142,36 +1141,18 @@ type stepResultChan struct {
 // Private interface
 // Add docs
 func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, stepID int, opts ...StepOption) (any, error) {
-	var wg sync.WaitGroup
-	stepResult := make(chan stepResultChan, 1)
-
-	wg.Add(1)
-
+	result := make(chan stepResultChan, 1)
 	go func() {
-		defer wg.Done()
 		res, err := c.RunAsStep(ctx, fn, c.WithNextStepID(stepID))
-		if err != nil {
-			stepRes := stepResultChan{
-				result: nil,
-				err:    err,
-			}
-			stepResult <- stepRes
-			return
-		}
-
-		stepRes := stepResultChan{
+		result <- stepResultChan{
 			result: res,
-			err:    nil,
+			err:    err,
 		}
-		stepResult <- stepRes
 	}()
 
-	wg.Wait()
-	close(stepResult)
-
-	res := <-stepResult
-
-	return res.result, res.err
+	resultChan := <-result
+	close(result)
+	return resultChan.result, resultChan.err
 }
 
 /****************************************/
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1091,11 +1091,7 @@ func (c *dbosContext) RunAsStep(_ DBOSContext, fn StepFunc, opts ...StepOption) 
 	return stepOutput, stepError
 }
 
-// how can I setup the sdk in a way to test it locally with other files?
-
-// Package level
-// Run step function using Go routines
-// Add odocs
+// TODO: Add docs --- will add once I get the implementation right
 func Go[R any](ctx DBOSContext, fn Step[R], opts ...StepOption) (R, error) {
 	if ctx == nil {
 		return *new(R), newStepExecutionError("", "", "ctx cannot be nil")
@@ -1133,13 +1129,13 @@ func Go[R any](ctx DBOSContext, fn Step[R], opts ...StepOption) (R, error) {
 	return typedResult, err
 }
 
+// TODO: move type above -- keeping it here for in case I need to modify quickly
 type stepResultChan struct {
 	result any
 	err    error
 }
 
-// Private interface
-// Add docs
+// TODO: Add docs --- will add once I get the implementation right
 func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, stepID int, opts ...StepOption) (any, error) {
 	result := make(chan stepResultChan, 1)
 	go func() {
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1117,11 +1117,10 @@ func Go[R any](ctx DBOSContext, fn Step[R], opts ...StepOption) (R, error) {
 	opts = append(opts, WithNextStepID(stepID))
 
 	// Type-erase the function
-
 	typeErasedFn := StepFunc(func(ctx context.Context) (any, error) { return fn(ctx) })
 
 	// run step inside a Go routine by passing stepID
-	result, err := ctx.Go(ctx, typeErasedFn, stepID, opts...)
+	result, err := ctx.Go(ctx, typeErasedFn, opts...)
 
 	// Step function could return a nil result
 	if result == nil {
@@ -1142,7 +1141,7 @@ type stepResultChan struct {
 }
 
 // TODO: Add docs --- will add once I get the implementation right
-func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, stepID int, opts ...StepOption) (any, error) {
+func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, opts ...StepOption) (any, error) {
 	result := make(chan stepResultChan, 1)
 	go func() {
 		res, err := c.RunAsStep(ctx, fn, opts...)
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"reflect"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -838,6 +839,7 @@ type StepOptions struct {
 	baseInterval  time.Duration // Initial delay between retries (default: 100ms)
 	maxInterval   time.Duration // Maximum delay between retries (default: 5s)
 	stepName      string        // Custom name for the step (defaults to function name)
+	stepID        int
 }
 
 // setDefaults applies default values to stepOptions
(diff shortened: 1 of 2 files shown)


af-md:
@maxdml does this feature have any conflict with what Alex said about pre generating stepIDs: https://discord.com/channels/1156433345631232100/1166779411920597002/1413954852618244267

apoliakov:
> @maxdml does this feature have any conflict with what @apoliakov said about pre generating stepIDs: https://discord.com/channels/1156433345631232100/1166779411920597002/1413954852618244267
> 
> 
> 
> It makes sense to run steps inside Go routines - as they tend to be better performant compared to standard execution - however the users should be told to write their code to wait for a step to complete (committed into DB) and then move onto the next step? probably that's what you were thinking of anyway...

Ah... what I said was a comment on how Python works. Here we may have an opportunity to make it act differently. But Max or Peter will need to opine on that

maxdml:
@maxdml @apoliakov there is a small misunderstanding here. The problem we are solving with this PR is the non deterministic generation of stepIDs, resulting from the execution of steps in goroutines.

What this PR will do is to serialize the generation of step IDs from within the workflow. That way, step IDs will be generated deterministically, before the `RunAsStep` code executes, and regardless of the order in which they complete, they'll always have the same stepID.

dbos/workflow.go
```diff
@@ -935,6 +936,18 @@ func WithMaxInterval(interval time.Duration) StepOption {
...
 }
 
+func WithNextStepID(stepID int) StepOption {
+	return func(opts *StepOptions) {
+		opts.preGeneratedStepID = &stepID
+	}
+}
+
+// StepOutcome holds the result and error from a step execution
+type stepOutcome[R any] struct {
```
maxdml:
```suggestion
type StepOutcome[R any] struct {
```

This should be made public.

Also let's add a bit more description in the comment as to where it is used, so `pkg.go.dev` picks it up.

dbos/workflows_test.go
```diff
@@ -569,6 +574,94 @@ func TestSteps(t *testing.T) {
...
+		assert.Equal(t, numSteps, len(results), "expected %d results, got %d", numSteps, len(results))
+		assert.Equal(t, 0, len(errors), "expected no errors, got %d", len(errors))
+
+		// Test step IDs are deterministic and in the order of execution
+		steps, err := GetWorkflowSteps(dbosCtx, handle.GetWorkflowID())
+		require.NoError(t, err, "failed to get workflow steps")
+		require.Len(t, steps, numSteps, "expected %d steps, got %d", numSteps, len(steps))
+		for i := 0; i < numSteps; i++ {
+			assert.Equal(t, i, steps[i].StepID, "expected step ID to be %d, got %d", i, steps[i].StepID)
+		}
```
maxdml: This does not test what you think it does: `GetWorkflowSteps` returns all the workflow steps sorted by ascending step ID, so you're testing the SQL, not the step ID attribution.

The way to test this would be to have each step take their ID as input and return their ID. Then, you can iterate over the channels and make sure that the iterator number == the step result from the channel.

* Channels should be ordered by stepID
* If the correct ID was attributed to the step, the step return value will be equal to the channel iterator

maxdml: In a second part, we should also exercise the recovery part, either by running `recoverPendingWorkflow`, or simply by executing the workflow again with the same workflowID. If there was a non determinism step attribution, DBOS would throw an error during the second `RunAsStep` -- which it shouldn't. 

For this to happen, however, we must ensure the workflow stays PENDING and does not return in the first, run, which we can achieve with an event (see [this example](https://github.com/dbos-inc/dbos-transact-golang/blob/e79823b16dd37a220e76500f1dac015bf46248db/dbos/workflows_test.go#L1160)). (If we don't do that, re-running the workflow will just get the workflow outcome, rather than going through the steps again.

dbos/workflow.go
```diff
@@ -1197,6 +1220,91 @@ func (c *dbosContext) RunAsStep(_ DBOSContext, fn StepFunc, opts ...StepOption)
 	return stepOutput, stepError
 }
 
+// Go runs a step inside a Go routine and returns a channel to receive the result.
+// Go generates a deterministic step ID for the step before running the step in a routine, since routines are not deterministic.
+// The step ID is used to track the steps within the same workflow and use the step ID to perform recovery.
+// The folliwing examples shows how to use Go:
```
maxdml, line 1275:
```suggestion
// The following example shows how to use Go:
```

dbos/workflow.go
```diff
@@ -1197,6 +1220,91 @@ func (c *dbosContext) RunAsStep(_ DBOSContext, fn StepFunc, opts ...StepOption)
...
+		result: typedResult,
+		err:    nil,
+	}
+
+	return outcomeChan, nil
+}
+
+func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, opts ...StepOption) (chan StepOutcome[any], error) {
+	// create a determistic step ID
+	stepName := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
```
maxdml, line 1336: This is not the step name we want to display: it is the name of the typed-erase step. We can retrieve the user-provided function's name by inspecting the options

```golang
	// Process functional options
	stepOpts := &stepOptions{}
	for _, opt := range opts {
		opt(stepOpts)
	}
name := stepOpts.stepName
```

dbos/workflow.go
```diff
@@ -1197,6 +1220,91 @@ func (c *dbosContext) RunAsStep(_ DBOSContext, fn StepFunc, opts ...StepOption)
...
+	opts = append(opts, WithStepName(stepName))
+
+	// Type-erase the function
+	typeErasedFn := StepFunc(func(ctx context.Context) (any, error) { return fn(ctx) })
+
+	result, err := ctx.Go(ctx, typeErasedFn, opts...)
+	// Step function could return a nil result
+	if result == nil {
+		return *new(chan StepOutcome[R]), err
+	}
```
maxdml, line 1306: we need to close the `result` chan. Let's do it after reading the outcome from it.

dbos/workflow.go
```diff
@@ -1197,6 +1220,91 @@ func (c *dbosContext) RunAsStep(_ DBOSContext, fn StepFunc, opts ...StepOption)
...
+
+	// Otherwise type-check and cast the result
+	typedResult, ok := outcome.result.(R)
+	if !ok {
+		return *new(chan StepOutcome[R]), fmt.Errorf("unexpected result type: expected %T, got %T", *new(R), result)
+	}
+	outcomeChan <- StepOutcome[R]{
+		result: typedResult,
+		err:    nil,
+	}
```
maxdml, line 1329: We're going to have to modify this after #175 is merged. Specifically, it is possible that the step output was returned from the database, encoded, in which case we'll need to decode it. We can mimick the code in `RunAsStep` that does the same.

dbos/workflows_test.go
```diff
@@ -855,6 +861,193 @@ func TestSteps(t *testing.T) {
...
+				if err != nil {
+					return "", err
+				}
+				resultChans = append(resultChans, resultChan)
+			}
+
+			for i, resultChan := range resultChans {
+				result1 := <-resultChan
+				if result1.err != nil {
+					errors <- result1.result.Error
```
maxdml, line 962: Why are we piping `result1.result.Error`? Is it every set to anything? If not let's just remove it from the result struct to clarify the code

dbos/workflows_test.go
```diff
@@ -855,6 +861,193 @@ func TestSteps(t *testing.T) {
...
+			}
+			return "", nil
+		}
+
+		RegisterWorkflow(dbosCtx, goWorkflow)
+		handle, err := RunWorkflow(dbosCtx, goWorkflow, "test-input")
+		require.NoError(t, err, "failed to run go workflow")
+		_, err = handle.GetResult()
+		close(results)
+		close(errors)
```
maxdml, line 975: nit: defer the closing after creation

dbos/workflow.go
maxdml: also in case of error just return outcomeChan, not a nil channel (which reading from blocks forever). That's a bit of a footgun for the user.

dbos/workflows_test.go
```diff
@@ -855,6 +861,193 @@ func TestSteps(t *testing.T) {
...
+		Error:  nil,
+	}, nil
+}
+
+// blocks indefinitely
+func stepThatBlocks(_ context.Context) (string, error) {
+	stepDeterminismStartEvent.Set()
+	fmt.Println("stepThatBlocks: started to block")
+	stepDeterminismEvent.Wait()
+	fmt.Println("stepThatBlocks: unblocked")
```
maxdml, line 894: remove print statements pls

dbos/workflows_test.go
```diff
@@ -855,6 +861,193 @@ func TestSteps(t *testing.T) {
...
+		require.NoError(t, err, "failed to run go workflow")
+		
+		// Wait for the first workflow to reach the blocking step
+		stepDeterminismStartEvent.Wait()
+		stepDeterminismStartEvent.Clear()
+
+		// Run the second workflow
+		handle2, err := RunWorkflow(dbosCtx, goWorkflow, "test-input", WithWorkflowID(handle.GetWorkflowID()))
+		
+		// If it throws an error, it's because of steps not being deterministically executed when using Go routines in the first workflow
```
maxdml, line 1034: The comment is not exactly accurate: determinism errors would come from `handle2.GetResult()`

maxdml:
Taking over in #221

PR #115, opened 2025-09-08
test dbos mocks

maxdml:
- Add a test exercising the mocking of the DBOSContext and WorkflowHandle interfaces
- We need a wrapper handle for the mock Path in `RunWorkflow` (it does return a new handle from the `DBOSContext.RunWorkflow`). This will only be used on the mocking path.


go.mod
```diff
@@ -67,3 +68,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	gotest.tools/v3 v3.5.2 // indirect
 )
+
+replace github.com/dbos-inc/dbos-transact-golang/dbos/mocks => ./mocks
```
maxdml: The only importers of this module are our tests and they should always get the mocks from the development directory.

dbos/workflow.go
```diff
@@ -216,6 +216,35 @@ func (h *workflowPollingHandle[R]) GetResult() (R, error) {
 	return *new(R), err
 }
 
+// Wrapper handle -- useful for handling mocks in RunWorkflow
```
kraftp, line 219: What does this do? Why do we need it? Do users need to worry about it?

go.mod
kraftp: What does this do? It looks sketchy.

dbos/mocks_test.go
```diff
@@ -0,0 +1,202 @@
+package dbos_test
```
kraftp, line 1: Do we need to commit the two mocks files?

go.mod
kraftp: I see, the mocks files are being committed as part of our core codebase--do we really need to do that?

dbos/workflow.go
maxdml: The situation we have with RunWorkflow, is that only the interface methods are mockable. If you want to mock RunWorkflow and have it return a mock handle, only the "inner" call with get a mock. Then, in the package level method, we use information from that handle to return a new typed handle. This new handle, not being the mock, makes testing impossible.

What this does  -- users don't need to worry -- is to fall back on a "proxy" handler if the interface RunWorkflow does not return the "normal" handlers (direct or polling). The proxy just calls in whatever was returned by RunWorkflow. This allow a mock handle to be passed through.

dbos/mocks_test.go
maxdml: You mean the mocks in `mocks`? Checking them will allow us to add a rule in GHA that'll generate the mocks and fail if there is a diff with what's checked-in. The goal is to help us detect as fast as possible (potentially breaking) changes to the interface, before merging in main.

kraftp: What does that actually accomplish? It will just trigger on every interface change, breaking or not. Lots of noise. I don't like having the mocks checked in, especially because it requires that weird rule in `go.mod`.

dbos/workflow.go
kraftp: Can't we just return a polling handle (passing in the mocked context) instead of this new thing?

maxdml: Returning a new polling handle would be returning a new object that is not the mock returned by `DBOSContext.RunWorkflow`. We need a way to pass through the mock while respecting the function signature (return a generic `WorkflowHandle[R]` interface). Unfortunately we cannot just "cast" the mock (`SomeMockType[any]`) to `WorkflowHandle[R]`: `any` and `R` are different type. This proxy type:
- Let us pass through the mock
- Let us returned a typed mock

Issue #86, opened 2025-08-29
Recv should handle context cancellation


maxdml:
Currently receive only respect its timeout, but does not consider `DBOSContext` cancellation.

PR #116, opened 2025-09-09
Fix comms durable sleeps

maxdml:
- Fix a bug where sleep steps ID were not always generated when `recv`/`getEvent` skipped sleeping (e.g., a value was already found, during recovery)
- Fix a bug were `recv` would return an error if it recorded a timeout value
- Handle context cancellation during `recv` (addresses #86)

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1817,7 +1817,7 @@ func (s *sysDB) recv(ctx context.Context, input recvInput) (any, error) {
 
 	// Deserialize the message
 	var message any
-	if messageString != nil { // nil message should never happen because they'd cause an error on the send() path
+	if messageString != nil { // nil message can happen on the timeout path only
 		message, err = deserialize(messageString)
 		if err != nil {
 			return nil, fmt.Errorf("failed to deserialize message: %w", err)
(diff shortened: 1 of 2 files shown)


dbos/system_database.go
```diff
@@ -2003,22 +2014,20 @@ func (s *sysDB) getEvent(ctx context.Context, input getEventInput) (any, error)
...
 
 		// Query the database again after waiting
 		row = s.pool.QueryRow(ctx, query, input.TargetWorkflowID, input.Key)
 		err = row.Scan(&valueString)
-		if err != nil {
-			if err == pgx.ErrNoRows {
-				value = nil // Event still doesn't exist
-			} else {
-				return nil, fmt.Errorf("failed to query workflow event after wait: %w", err)
-			}
```
maxdml, line 2017: simplify: the zero value of an `any` value is already `nil`

dbos/system_database.go
```diff
@@ -1772,6 +1778,20 @@ func (s *sysDB) recv(ctx context.Context, input recvInput) (any, error) {
...
+	} else {
+		// Do a "zero" sleep to record the sleep step in the workflow history
+		// This is to handle cases were consecutive Recv calls for the same key are made
+		// The key would exist and without this, the step IDs would be: 0 recv, 1 sleep, 2 recv, 4 recv, etc
+		_, err := s.sleep(ctx, sleepInput{
+			duration: 0,
+			stepID:   &sleepStepID,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to sleep before recv timeout: %w", err)
```
maxdml: I am moderately happy about this, but its the simpler and clearer: just always add a sleep step to receive calls, even if the key exists.

Otherwise we can have this kind of state:

```sql
function_id | function_name
1 | DBOS.sleep 
0 | DBOS.recv
2 | DBOS.recv
4 | DBOS.recv 
```

For a program that does consecutive receives on the same key.

Other solutions to not always do the "zero sleep" requires detecting this pattern or that we are recovering

PR #120, opened 2025-09-10
Handle DBOS operations as workflow steps

maxdml:
Call DBOS operations inside `RunAsStep` if within a workflow.

One thing this PR had to fix is the forked workflow ID generation, which was at a higher level and not idempotent. The systemDB forkWorkflow now returns the forked workflow ID.


dbos/queues_test.go
```diff
@@ -926,23 +926,17 @@ func TestQueueTimeouts(t *testing.T) {
...
 		require.ErrorAs(t, err, &dbosErr, "expected error to be of type *DBOSError, got %T", err)
 		assert.Equal(t, AwaitedWorkflowCancelled, dbosErr.Code, "expected error code to be AwaitedWorkflowCancelled")
-
-		// enqueud workflow should have been cancelled
-		status, err := handle.GetStatus()
-		require.NoError(t, err, "failed to get status of enqueued workflow")
-		assert.Equal(t, WorkflowStatusCancelled, status.Status, "expected enqueued workflow status to be WorkflowStatusCancelled")
-
-		return "should-never-see-this", nil
+		return "", nil
```
maxdml, line 939: Now, `GetStatus` from within a workflow that was cancelled will return an error -- because the step checkpointing detects the workflow was cancelled.

dbos/system_database.go
```diff
@@ -546,6 +547,45 @@ type listWorkflowsDBInput struct {
 
 // ListWorkflows retrieves a list of workflows based on the provided filters
 func (s *sysDB) listWorkflows(ctx context.Context, input listWorkflowsDBInput) ([]WorkflowStatus, error) {
+	functionName := "DBOS.listWorkflows"
```
kraftp: There's no way to factor these huge intricate code blocks out as was done in Python/TS? Why not literally call these methods with runStep?

maxdml: I can look into `RunAsStep`! I didn't want a `run_as_special_step` wrapper

PR #127, opened 2025-09-16
DBOS client interface

maxdml:


dbos/client.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// Register the input and outputs for gob encoding
+	var typedInput P
+	gob.Register(typedInput)
+	var typedOutput R
+	gob.Register(typedOutput)
+
+	// Call the interface method with the same signature
+	handle, err := c.Enqueue(queueName, workflowName, input, opts...)
```
kraftp, line 236: There's no way to avoid still needing this in the public interface?

kraftp (approved):

PR #133, opened 2025-09-17
make sys db schema configurable

maxdml:
Users can now provide a `DatabaseSchema` to `Config` and `ClientConfig`

### DIFF
--- a/dbos/dbos_test.go
+++ b/dbos/dbos_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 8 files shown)


dbos/system_database.go
```diff
@@ -150,7 +148,28 @@ const (
 	_DB_RETRY_INTERVAL               = 1 * time.Second
 )
 
-func runMigrations(databaseURL string) error {
+func runMigrations(databaseURL string, schema string) error {
+	// Process the migration template with the schema
+	tmpl, err := template.New("migration").Parse(migration1Template)
```
kraftp: This could be simpler, but it's not a huge deal

dbos/dbos_test.go
```diff
@@ -325,4 +326,111 @@ func TestConfig(t *testing.T) {
...
+		sysDB, ok := dbosCtx.systemDB.(*sysDB)
+		require.True(t, ok, "expected sysDB")
+
+		// Verify schema name was set correctly
+		assert.Equal(t, customSchema, sysDB.schema, "schema name should match custom schema")
+
+		// Verify all expected tables exist in the custom schema
+		dbCtx := context.Background()
+
+		// Test workflow_status table in custom schema
```
kraftp, line 369: We should try actually doing something with DBOS and this custom schema instead of just checking a bunch of tables exist

dbos/system_database.go
maxdml: You mean something like calling `SET search_path TO custom_schema, dbos, public;` and not having any schema name in the migration SQL ?

kraftp: That's not really simpler--I was thinking just using Sprintf on a migration string literal instead of all this parsing and templating.

dbos/dbos_test.go
maxdml: We do it in https://github.com/dbos-inc/dbos-transact-golang/pull/133/files#diff-05899e4d3d85f6b02a7e4f3668bb0facf6fd4b9fbed89a1c6097b2e712e10caf, albeit that's not exercising all the system db methods.

dbos/system_database.go
maxdml: Yeah that's fair, I'm not a big fan of `Sprintf` and its a bit clunky (specifically having to get the count of substituted variables right) but that's simpler.

dbos/dbos_test.go
kraftp: Let's add a dedicated custom schema test that actually exercises the system database--particularly send and recv, since the triggers there are the things most likely to break.

maxdml: I'm happy to add a dedicated test that exercises all the system db methods, but, may I ask, why are the triggers special? `pg_notify` works across schema

kraftp: Because they're intricate, easy to get wrong

kraftp (approved):

Issue #94, opened 2025-08-29
Resilient notification listener


maxdml:
The notification listener used by `Recv` / `GetEvent` is not resilient to connections issues. If the connection fails, the listener should be restarted.

PR #134, opened 2025-09-17
use custom pool for everything in systemp db

maxdml:
This PR:
- Ensure we use the user pgxpool for every interaction with the system DB, if provided
- Moves the notification listener to using a connection from from said pool, rather than using a standalone connection
- Make the notification listener loop resilient to connection failures (addresses #94)

Note that `OnNotification` has to be set before a connection is started (which is too late once Acquired from the pool), which is why the logic to handle notifications is now inlined in the listener.

### DIFF
--- a/dbos/dbos_test.go
+++ b/dbos/dbos_test.go
@@ -506,7 +506,7 @@ func TestCustomPool(t *testing.T) {
 		return receivedMsg, nil
 	}
 
-	t.Run("NewSystemDatabaseWithCustomPool", func(t *testing.T) {
+	t.Run("CustomPool", func(t *testing.T) {
 		// Custom Pool
 		databaseURL := getDatabaseURL()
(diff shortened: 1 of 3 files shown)


dbos/dbos_test.go
```diff
@@ -517,3 +461,82 @@ func TestCustomSystemDBSchema(t *testing.T) {
...
+
+		RegisterWorkflow(dbosCtx, wf)
+
+		// Launch the DBOS context
+		err = dbosCtx.Launch()
+		require.NoError(t, err)
+		defer dbosCtx.Shutdown(1 * time.Minute)
+
+		// Run a workflow
+		_, err = RunWorkflow(dbosCtx, wf, "test-input")
```
kraftp, line 636: This should also do send/recv to verify the listener-with-custom-pool stuff works.

PR #139, opened 2025-09-22
Handle conflicting executions

maxdml:
If a workflow observes the returned error is a `DBOSError` with code `ConflictingIDError`, fall back to `awaitWorkflowResult` in the workflow goroutine.

Note that we cannot distinguish this error if the user handles it (e.g., during `setEvent`) and wraps it with some other error message.

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1189,9 +1189,8 @@ func (s *sysDB) recordOperationResult(ctx context.Context, input recordOperation
 		return fmt.Errorf("failed to serialize output: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


dbos/system_database.go
```diff
@@ -1222,10 +1215,6 @@ func (s *sysDB) recordOperationResult(ctx context.Context, input recordOperation
 		return err
 	}
 
-	if commandTag.RowsAffected() == 0 {
-		s.logger.Warn("RecordOperationResult No rows were affected by the insert")
-	}
-
```
maxdml, line 1228: This should never happen

PR #144, opened 2025-09-23
Fix: Workflow Identity

### LINKED ISSUE #92: Workflow identity
Transact supports identifying workflows and RBAC through `authenticated_user` and `authenticated_roles`.

This can be done by adding a few functional options to `RunWorkflow` and `EnqueueWorkflow`:
- `WithAuthenticatedUser`
- `WithAuthenticatedRoles`
- `WithAssumedRole`

vr-varad:
Fixes #92


dbos/system_database.go
```diff
@@ -489,7 +490,7 @@ func (s *sysDB) insertWorkflowStatus(ctx context.Context, input insertWorkflowSt
 		input.status.QueueName,
 		input.status.AuthenticatedUser,
 		input.status.AssumedRole,
-		input.status.AuthenticatedRoles,
+		pgtype.Array[string]{Elements: input.status.AuthenticatedRoles},
```
vr-varad: As the authenticated roles is an array of string, I was getting this error for the news tests
```
failed to insert workflow status: failed to encode args[6]: unable to encode []string{"reader", "writer"} into text format for text (OID 25): cannot find encode plan
```
But then added this to allow inserting array, which is giving bunch of error, i tried but coudn't get another way.
What are your thoughts?

vr-varad: @maxdml @kraftp

maxdml: The reason you are having this issue is that `authenticated_roles` is represented as a single `TEXT` column in Postgres. To match the other DBOS implementations, we should marshal `input.status.AuthenticatedRoles` to JSON and insert that in the database. On the read path, we should unmarshal it.

dbos/system_database.go
```diff
@@ -483,14 +483,18 @@ func (s *sysDB) insertWorkflowStatus(ctx context.Context, input insertWorkflowSt
 	var result insertWorkflowResult
 	var timeoutMSResult *int64
 	var workflowDeadlineEpochMS *int64
+
+	// Marshal authenticated roles (slice of strings) to JSON for TEXT column
+	authenticatedRoles, _ := json.Marshal(input.status.AuthenticatedRoles)
```
maxdml: Please handle errors explicitly

maxdml (approved):
lgtm, thanks @vr-varad

PR #146, opened 2025-09-23
Db retries

maxdml:
Address https://github.com/dbos-inc/dbos-transact-golang/issues/93

This PR adds a `retry` method that can take in another function for retry until a specific condition is met. `retry` can be configured with functional options and has a `retryWithResult` generic counterpart for functions returning a value.

All system database methods are now called within `retry` or `retryWithResult`.

For example

```go
dequeuedWorkflows, err := retryWithResult(ctx, func() ([]dequeuedWorkflow, error) {
return ctx.systemDB.dequeueWorkflows(ctx, dequeueWorkflowsInput{
	queue:              queue,
	executorID:         ctx.executorID,
	applicationVersion: ctx.applicationVersion,
})
}, withRetrierLogger(qr.logger))
```

retry performs exponential backoff with jitter. It defaults to infinite retries.

The default condition for retrying is `isRetryablePGError`,  which matches the function returned error with a set of connection errors (postgres codes, pgx connection errors, `net.Error`)

Note: in the particular case of RunWorkflow, which does manage transactions, the entire function has to be retried. This is because the transaction object becomes invalid as soon as Commit/Rollback have been called, regardless of whether the operation was successful. See https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/tx.go#L180

This PR also improves `DBOSError` with the ability to unwrap the underlying error, if any.

### DIFF
--- a/cmd/dbos/templates/dbos-toolbox/go.mod.tmpl
+++ b/cmd/dbos/templates/dbos-toolbox/go.mod.tmpl
@@ -1,3 +1,5 @@
 module {{.ProjectName}}
 
 go 1.22.0
+
+toolchain go1.25.0


cmd/dbos/templates/dbos-toolbox/go.mod.tmpl
```diff
@@ -1,5 +1,3 @@
 module {{.ProjectName}}
 
-go 1.22.0
-
-toolchain go1.25.0
```
maxdml, line 5: Not needed for template

dbos/errors.go
```diff
@@ -28,7 +28,6 @@ const (
 type DBOSError struct {
 	Message string        // Human-readable error message
 	Code    DBOSErrorCode // Error type code for programmatic handling
-	IsBase  bool          // Internal errors that shouldn't be caught by user code
```
maxdml, line 31: unused

.github/workflows/tests.yml
```diff
@@ -44,16 +44,6 @@ jobs:
...
-    - name: Cache Go modules
-      uses: actions/cache@v4
-      with:
-        path: |
-          ~/go/pkg/mod
-          ~/.cache/go-build
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
-
```
maxdml, line 56: Doesn't seem very useful

cmd/dbos/postgres.go
```diff
@@ -174,7 +175,7 @@ func startDockerPostgres() error {
 		return err
 	}
 
-	logger.Info("Postgres available", "url", fmt.Sprintf("postgres://postgres:%s@localhost:5432", password))
+	logger.Info("Postgres available", "url", fmt.Sprintf("postgres://postgres:%s@localhost:5432", url.QueryEscape(password)))
```
maxdml, line 187: fix

cmd/dbos/postgres.go
```diff
@@ -156,6 +163,9 @@ func startDockerPostgres() error {
 			},
 		},
 		AutoRemove: true,
+		Binds: []string{
+			fmt.Sprintf("%s:%s", hostPgDataVolumeName, pgData),
+		},
```
maxdml, line 168: It turns out that an explicit volume binding is required. On MacOS, exporting `PGDATA` is enough to instruct Docker desktop to mount the right thing. Docker desktop uses a light VM to run postgres, and persist this volume across container restarts.

On the github action runner -- and likely on other environments too? The mount is not persisted across containers restarts.

dbos/workflow.go
```diff
@@ -437,30 +450,20 @@ func RegisterWorkflow[P any, R any](ctx DBOSContext, fn Workflow[P, R], opts ...
...
-				return nil, fmt.Errorf("failed to get workflow ID: %w", err)
-			}
-			err = ctx.(*dbosContext).systemDB.updateWorkflowOutcome(WithoutCancel(ctx), updateWorkflowOutcomeDBInput{
-				workflowID: wfID,
-				status:     WorkflowStatusError,
-				err:        newWorkflowUnexpectedInputType(fqn, fmt.Sprintf("%T", typedInput), fmt.Sprintf("%T", input)),
-			})
-			if err != nil {
-				return nil, fmt.Errorf("failed to record unexpected input type error: %w", err)
-			}
```
maxdml, line 451: This is not required because `typedErasedWorkflow` is always call within `RunWorkflow`, in a goroutine that update the workflow outcome.

integration/mocks_test.go
```diff
@@ -190,9 +190,4 @@ func TestMocks(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
-	mockCtx.AssertExpectations(t)
-	mockChildHandle.AssertExpectations(t)
-	mockGenericHandle.AssertExpectations(t)
```
maxdml, line 196: A mock built with `mocks.NewMoc...` already checks expectations before destroying the object.

dbos/workflow.go
```diff
@@ -578,7 +581,14 @@ func RunWorkflow[P any, R any](ctx DBOSContext, fn Workflow[P, R], input P, opts
...
 	})
 
-	handle, err := ctx.RunWorkflow(ctx, typedErasedWorkflow, input, opts...)
+	// Wrap the RunWorkflow call with retryWithResult for database operation retries
+	var logger *slog.Logger
+	if dbosCtx, ok := ctx.(*dbosContext); ok {
+		logger = dbosCtx.logger
+	}
+	handle, err := retryWithResult(ctx, func() (WorkflowHandle[any], error) {
+		return ctx.RunWorkflow(ctx, typedErasedWorkflow, input, opts...)
```
kraftp: I don't get this. The idea is to retry individual database operations in a workflow if they fail due to a database disconnection error, not to retry the entire workflow.

maxdml: The challenge is that if the transaction Commit or Rollback fails during a connection error, the entire transaction is unusable. We have to redo all its operations. This is the simplest way to achieve this. RunWorkflow is entirely transactional so this is safe from a durability point of view.

Note that the actual workflow only starts after the transaction is committed.

maxdml: (specifically this is the issue: https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/tx.go#L190)

kraftp: I still don't follow. Of course we have to retry the transaction! But we should retry at the granularity of transactions/system database operations, not retry all of RunWorkflow.

maxdml: If this fails https://github.com/dbos-inc/dbos-transact-golang/blob/main/dbos/workflow.go#L820-L822

Then all the previous database operations must be retried. The other code that is not these operations is internal & idempotent.

maxdml: The transaction _object_ itself becomes unusable

kraftp: But this retries the ENTIRE WORKFLOW if I understand it correctly. Also, one system database call should equal one transaction, they shouldn't leak to the workflow level (and there's no added benefit from doing so).

maxdml: Retry is configured to only retry on DB errors. RunWorkflow can only return such error up to before the user code is started in a different goroutine. That is, retrying will never run the user-provided code twice.

The transaction in RunWorkflow encompasses:
- insert workflow status
- recording the parent - child relationship, if any

Beyond the transaction, retrying RunWorkflow will only repeat `checkChildWorkflow`.

We could wrap the transaction in a closure, but what's the advantage of that?

maxdml: I did find one quirk to my idempotency claim, which is that we increment the stepID if we retry. I moved the retry to wrap the transaction only

dbos/client.go
```diff
@@ -162,28 +162,32 @@ func (c *client) Enqueue(queueName, workflowName string, input any, opts ...Enqu
...
-	}
-	defer tx.Rollback(uncancellableCtx) // Rollback if not committed
-
-	// Insert workflow status with transaction
-	insertInput := insertWorkflowStatusDBInput{
-		status: status,
-		tx:     tx,
-	}
-	_, err = dbosCtx.systemDB.insertWorkflowStatus(uncancellableCtx, insertInput)
+	err := retry(dbosCtx, func() error {
```
kraftp: Why do we need retries here? It's the client, it can just fail, there's no deeper risk

dbos/workflow.go
```diff
@@ -734,92 +732,103 @@ func (c *dbosContext) RunWorkflow(_ DBOSContext, fn WorkflowFunc, input any, opt
...
-		maxRetries: params.maxRetries,
-		tx:         tx,
-	}
-	insertStatusResult, err := c.systemDB.insertWorkflowStatus(uncancellableCtx, insertInput)
-	if err != nil {
-		c.logger.Error("failed to insert workflow status", "error", err, "workflow_id", workflowID)
-		return nil, err
-	}
+	// Init status and record child workflow relationship in a single transaction
+	err := retry(c, func() error {
```
kraftp, line 739: These don't need to be done in one transaction. This transaction is too large and I'm not convinced retrying it is safe. We should get rid of this large transaction and just do init and `recordChildWorkflow` separately, as the other languages do.

maxdml: These are two back to back insert statements which really go together... what are the safety issues?

kraftp: That this is a giant block with a ton of stuff in it (sure, two inserts, but also lots of in-memory bookkeeping that doesn't belong to the transaction) that retries, and I don't know if it's safe--are there other side effects (maybe in memory) in that block?

dbos/client.go
maxdml: Not necessarily strictly speaking, but nice to have from a reliable library imo. Right now, all the client methods, which are wrapping the context methods, will retry.

Are you suggesting we should do the minimum amount of retries to avoid getting a corrupted Transact process (e.g., queue runner unusable), but should let the user handle connection errors on all other paths? E.g., cancel workflow.

dbos/workflow.go
maxdml: Got it 👍 . In addition to the SQL queries, this block does create the workflow context and set up the cancellation function. It shouldn't be part of the transaction (as there are no occasion to error in that logic). I'm moving it out!

dbos/client.go
kraftp: It's not really a huge deal either way--retries in the client are excessive, but not unsafe (and the Python client retries)

dbos/system_database.go
```diff
@@ -2547,3 +2553,219 @@ func maskPassword(dbURL string) (string, error) {
...
+
+// withRetrierLogger sets the logger for the retrier
+func withRetrierLogger(logger *slog.Logger) retryOption {
+	return func(c *retryConfig) {
+		c.logger = logger
+	}
+}
+
+// retry executes a function with retry logic using functional options
+func retry(ctx context.Context, fn func() error, options ...retryOption) error {
```
kraftp, line 2632: It's not a huge deal, but this thing is massively overengineered, given that it's always called with the same parameters. Somewhat confusing.

chaos_tests/chaos_test.go
```diff
@@ -0,0 +1,449 @@
...
+	var wg sync.WaitGroup
+	defer wg.Wait()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	PostgresChaosMonkey(t, ctx, &wg)
+
+	queue := dbos.NewWorkflowQueue(dbosCtx, "test_queue")
+
+	// Define step functions
+	stepOne := func(ctx dbos.DBOSContext, x int) (int, error) {
```
kraftp, line 407: The other thing we could check is if a scheduled workflow keeps ticking after reconnection

kraftp (approved):
Looks good, important tests and fixes!

dbos/system_database.go
maxdml: Removed the unused options

dbos/client.go
maxdml: I like the idea of doing less. We can easily add retries to client methods (enqueue, but also RetrieveWorkflow, CancelWorkflow, ResumeWorkflow, ListWorkflows, GetWorkflowSteps and ForkWorkflow) later if need be. The most important is to avoid having unusable-but-running Transact processes.

PR #155, opened 2025-09-26
Starter examples: replace dbos-toolbox with dbos-go-starter

ngnhng:
## Summary
Hi, thanks for the great project, I am currently trying to demo this my team but the current example sucked. So i went ahead and replace it with the dbos-go-starter.


cmd/dbos/templates/dbos-go-starter/main.go.tmpl
```diff
@@ -0,0 +1,155 @@
...
+/*****************************/
+/**** Main Function **********/
+/*****************************/
+
+func main() {
+	// Create DBOS context
+	var err error
+	dbosCtx, err = dbos.NewDBOSContext(context.Background(), dbos.Config{
+		DatabaseURL: os.Getenv("DBOS_SYSTEM_DATABASE_URL"),
+		AppName:     "dbos-toolbox",
```
maxdml:
```suggestion
		AppName: "{{.ProjectName}}",
```

cmd/dbos/init.go
```diff
@@ -70,6 +71,9 @@ func runInit(cmd *cobra.Command, args []string) error {
 
 		// Write output file
 		outputPath := filepath.Join(projectName, outputFile)
+		if err := os.MkdirAll(filepath.Dir(outputPath), 0755); err != nil {
+			return fmt.Errorf("failed to create directory for %s: %w", outputFile, err)
+		}
```
maxdml, line 76: Why is this needed? `os.Write` will create the folders is needed

n-g-u-y-e-n: in `templates` map, we have this entry:

```go
"templates/dbos-go-starter/app.html": "html/app.html",
```
when `outputFile` is "html/app.html", `filepath.Dir(outputPath)` will resolve to "dbos-go-starter/html". The `os.MkdirAll` call then ensures that the html subdirectory is created inside the main project directory before `os.WriteFile` attempts to create `app.html` within it

so without the `os.MkdirAll`, we will get:

```
./dbos init
Error: failed to write html/app.html: open dbos-go-starter/html/app.html: no such file or directory
```

Issue #150, opened 2025-09-24
Add interface/package methods to list workflow registry methods


maxdml:
Add a helper to list:
- The registered workflows
- The registered scheduled workflows

The helper should return the workflows registration parameter as well.

PR #157, opened 2025-09-28
helper to list workflow registration

PhakornKiong:
Adds exported methods to list registered workflows and scheduled workflows with their registration parameters.

Resolves https://github.com/dbos-inc/dbos-transact-golang/issues/150

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -321,6 +321,9 @@ func registerScheduledWorkflow(ctx DBOSContext, workflowName string, fn Workflow
 		panic("Cannot register scheduled workflow after DBOS has launched")
 	}
 
+	// Store the schedule information in the scheduled workflow registry
+	c.scheduledWorkflowRegistry.Store(workflowName, cronSchedule)
+
 	c.getWorkflowScheduler().Start()
 	var entryID cron.EntryID
 	entryID, err := c.getWorkflowScheduler().AddFunc(cronSchedule, func() {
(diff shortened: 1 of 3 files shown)


dbos/workflows_test.go
```diff
@@ -4131,3 +4131,156 @@ func TestSpecialSteps(t *testing.T) {
...
+		return "hourly", nil
+	}, WithSchedule("0 0 * * * *"))
+
+	// Stop the cron scheduler to prevent goroutine leaks in tests
+	// The scheduler is started when scheduled workflows are registered
+	if dbosCtx.(*dbosContext).workflowScheduler != nil {
+		ctx := dbosCtx.(*dbosContext).workflowScheduler.Stop()
+		<-ctx.Done() // Wait for it to stop
+		dbosCtx.(*dbosContext).workflowScheduler = nil
+	}
```
maxdml: We shouldn't have to do this: `setupDBOS` calls `Shutdown` on the context, which stops the scheduler.

dbos/workflows_test.go
```diff
@@ -4131,3 +4131,156 @@ func TestSpecialSteps(t *testing.T) {
...
+
+	t.Run("ListRegisteredWorkflows", func(t *testing.T) {
```
maxdml, line 4258: Let's just have the package level tests. The interface tests are redundant (the package functions are a very thin layer)

maxdml (commented):
Thanks for the PR, @PhakornKiong. This is going on a good direction!

I think we can just reuse the existing `workflowRegistryEntry`, make it public and add a `CronSchedule` property, instead of adding two new types.

This also means we can keep using a single map to store all the workflows. The list functions can filter in/out the scheduled workflows.

dbos/workflows_test.go
PhakornKiong: Agreed. I was getting the following error. Realized that I did not launch dbos. Just sharing here for future reference

```
time=2025-10-01T03:38:37.431+08:00 level=INFO msg="Initializing DBOS context" app_name=test-app dbos_version=(devel)
time=2025-10-01T03:38:37.432+08:00 level=INFO msg="Connecting to system database" database_url="postgres://postgres:***@localhost:5432/dbos?sslmode=disable" schema=dbos
time=2025-10-01T03:38:38.111+08:00 level=INFO msg="Registered scheduled workflow" fqn=github.com/dbos-inc/dbos-transact-golang/dbos.simpleWorkflowWithSchedule cron_schedule="0 0 * * * *"
time=2025-10-01T03:38:38.111+08:00 level=INFO msg="Cleaning up DBOS instance..."
--- FAIL: TestRegisteredWorkflowListing (1.28s)
   /dbos/dbos-transact-golang/dbos/utils_test.go:76: found unexpected goroutines:
        [Goroutine 50 in state select, with github.com/robfig/cron/v3.(*Cron).run on top of the stack:
        github.com/robfig/cron/v3.(*Cron).run(0xc00032a320)
        	/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:263 +0xdf0
        created by github.com/robfig/cron/v3.(*Cron).Start in goroutine 19
        	/go/pkg/mod/github.com/robfig/cron/v3@v3.0.1/cron.go:222 +0x100
        ]
FAIL
FAIL	github.com/dbos-inc/dbos-transact-golang/dbos	1.613s
```

PhakornKiong:
> Thanks for the PR, @PhakornKiong. This is going on a good direction!
> 
> I think we can just reuse the existing `workflowRegistryEntry`, make it public and add a `CronSchedule` property, instead of adding two new types.
> 
> This also means we can keep using a single map to store all the workflows. The list functions can filter in/out the scheduled workflows.

I've updated the PR with the following changes:
- Expose `WorkflowRegistryEntry`
- Add `FQN` to `WorkflowRegistryEntry`
- Implement filtering for `ListRegisteredWorkflows()` with functional option pattern

dbos/workflows_test.go
maxdml: Good catch. This must be because we do start the scheduler in `registerScheduledWorkflow`. We should do it in `Launch` only. I'll fix it!

dbos/queue.go
```diff
@@ -243,7 +243,7 @@ func (qr *queueRunner) run(ctx *dbosContext) {
 					}
 				}
 
-				_, err := registeredWorkflow.wrappedFunction(ctx, input, WithWorkflowID(workflow.id))
+				_, err := registeredWorkflow.WrappedFunction(ctx, input, WithWorkflowID(workflow.id))
```
maxdml: Let's keep the wrapped function private -- no need to expose it

PhakornKiong: Updated!

dbos/workflow.go
```diff
@@ -1951,3 +1964,49 @@ func GetWorkflowSteps(ctx DBOSContext, workflowID string) ([]StepInfo, error) {
...
+// WithScheduledOnly filters to only return scheduled workflows (those with a cron schedule).
+func WithScheduledOnly() ListRegisteredWorkflowsOption {
+	return func(p *listRegisteredWorkflowsOptions) {
+		p.scheduledOnly = true
+	}
+}
+
+// ListRegisteredWorkflows returns information about workflows registered with DBOS.
+// Each WorkflowRegistryEntry contains:
+// - WrappedFunction: The underlying workflow implementation
```
maxdml: nit: update go doc too

PR #160, opened 2025-09-29
Test db permissions

maxdml:


cmd/dbos/cli_integration_test.go
```diff
@@ -68,17 +66,26 @@ func TestCLIWorkflow(t *testing.T) {
...
 			name:       "DefaultSchema",
 			schemaName: "dbos",
-			schemaArgs: []string{}, // No schema argument, use default
+			dbRole:     "postgres",
+			args:       []string{},
+		},
+		{
+			name:       "FunnySchema",
+			schemaName: "F8nny_sCHem@-n@m3",
+			dbRole:     "notpostgres",
```
kraftp: Put some funny characters in the role to verify everything is quoted correctly

cmd/dbos/cli_integration_test.go
```diff
@@ -222,32 +261,59 @@ func testProjectInitialization(t *testing.T, cliPath string) {
...
+	args := append([]string{"--verbose", "migrate"}, baseArgs...)
+	if dbRole != "postgres" {
+		args = append(args, "--app-role", dbRole)
+	}
+	cmd := exec.Command(cliPath, args...)
+	cmd.Env = append(os.Environ(), "DBOS_SYSTEM_DATABASE_URL="+getDatabaseURL("postgres"))
+	output, err := cmd.CombinedOutput()
+	require.NoError(t, err, "Migrate command failed: %s", string(output))
+	assert.Contains(t, string(output), "DBOS migrations completed successfully", "Output should confirm migration")
+	fmt.Println(string(output))
```
kraftp, line 295: We also need to test the app actually works with the new role after the migration

maxdml: The other tests do that (they'll use the role that's passed to the suite)

kraftp (approved):

PR #161, opened 2025-09-30
Fix: Register WorkflowHandle for gob encoding automatically

vr-varad:


dbos/dbos_test.go
```diff
@@ -719,3 +719,38 @@ func TestCustomPool(t *testing.T) {
...
+
+	RegisterWorkflow(ctx, wf)
+
+	// Launch a workflow
+	err = Launch(ctx)
+	require.NoError(t, err)
+	defer Shutdown(ctx, 1*time.Minute)
+
+	// Run a workflow
+	handle, err := RunWorkflow(ctx, wf, "test-input")
```
vr-varad: I hope this is what is required to test the registration for the workflow-handlers

vr-varad: // @maxdml

maxdml: What we need is a workflow that returns a `WorkflowHandle`

For example

```go
childWf  := func(ctx DBOSContext, input string) (string, error) {
	return input, nil
}

wf := func(ctx DBOSContext, input string) (WorkflowHandle, error) {
	return RunWorkflow(ctx, childWf, input)
}
```

Without the registration you did above, this should fail at runtime. We should make the example above a bit more complicated, so it also cover steps (e.g., have `wf` call a step, after starting the child workflow, that takes in the child's handle and return it, then have `wf` return the handle)

Then, when you call `GetResult` on the workflow handle, the result should be a valid `WorkflowHandle` which you can use to `GetResult`, `GetStatus`, etc

maxdml:
> Does the failing test mean the registration has failed? // @maxdml

I think there is another, unrelated bug. It looks like we don't check if the concrete value provided to `safeGobRegister` is nil. Gob does panic if that's the case. I'll be providing a fix soon.

maxdml:
@vr-varad the issue is that we can't automatically register interfaces during RegisterWorkflow. In this case, when you try to register a workflow that returns an interface, behind the hood we create a zero value and attempt to register it. For all zero values, this works fine, but for interfaces, [gob requires that the underlying concrete type has been registered first](https://pkg.go.dev/encoding/gob#example-package-Interface).

There is another issue I now realize: the workflow handles don't export any field, which makes them unsuitable for encoding as well.

In this case let's just cancel this feature, ignore registration of nil interfaces, and ask users to manually register their types.

vr-varad:
Makes Sense @maxdml!!

Issue #85, opened 2025-08-29
Cancellable GetResult


maxdml:
On polling handles, GetResult is current blocking. It should:
- Respect the handle's `DBOSContext` cancellation
- Accept a timeout

This feature implies making the underlying `awaitWorkflowResult` cancellable.

PhakornKiong:
@maxdml Would love to give it a shot if thats cool.

pranshu-raj-211:
@PhakornKiong please go ahead, my implementation had some issues I'm not able to figure out.

PR #169, opened 2025-10-04
Cancellable get results

PhakornKiong:
Resolves https://github.com/dbos-inc/dbos-transact-golang/issues/85

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -197,9 +197,9 @@ func (h *workflowHandle[R]) GetResult(opts ...GetResultOption) (R, error) {
 		}
 		return h.processOutcome(outcome)
 	case <-h.dbosContext.Done():
-		return *new(R), h.dbosContext.Err()
+		return *new(R), context.Cause(h.dbosContext)
 	case <-timeoutChan:
-		return *new(R), fmt.Errorf("workflow result timeout after %v", options.timeout)
+		return *new(R), fmt.Errorf("workflow result timeout after %v: %w", options.timeout, context.DeadlineExceeded)
 	}
 }
 
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -35,7 +35,7 @@ func simpleWorkflowWithStep(dbosCtx DBOSContext, input string) (string, error) {
 }
 
 func slowWorkflow(dbosCtx DBOSContext, sleepTime time.Duration) (string, error) {
-	time.Sleep(sleepTime)
+	Sleep(dbosCtx, sleepTime)
 	return "done", nil
 }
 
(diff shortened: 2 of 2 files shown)


dbos/workflow.go
```diff
@@ -162,12 +178,33 @@ type workflowHandle[R any] struct {
...
+
+	select {
+	case outcome, ok := <-h.outcomeChan:
+		if !ok {
+			// Return error if channel closed (happens when GetResult() called twice)
+			return *new(R), errors.New("workflow result channel is already closed. Did you call GetResult() twice on the same workflow handle?")
+		}
+		return h.processOutcome(outcome)
+	case <-h.dbosContext.Done():
+		return *new(R), h.dbosContext.Err()
```
PhakornKiong: @maxdml I'm unable to get to this code path, so not sure if we really need it.

Tried shutting down DBOS server, but it get DB error from `h.outcomeChan` instead.

Let me know your thoughts

dbos/workflows_test.go
```diff
@@ -4523,3 +4528,82 @@ func TestWorkflowIdentity(t *testing.T) {
...
+
+	t.Run("WorkflowHandleNoTimeout", func(t *testing.T) {
+		// Test without timeout - should work normally
+		handle, err := RunWorkflow(dbosCtx, slowWorkflow, 1*time.Millisecond)
+		require.NoError(t, err, "failed to start workflow")
+
+		result, err := handle.GetResult()
+		require.NoError(t, err, "GetResult without timeout should succeed")
+		assert.Equal(t, "done", result)
+	})
```
maxdml: No need this test -- we exercise this path plenty already

dbos/workflows_test.go
```diff
@@ -4523,3 +4528,82 @@ func TestWorkflowIdentity(t *testing.T) {
...
+		assert.Equal(t, "done", result1)
+
+		// Sleep briefly to ensure channel is closed
+		time.Sleep(10 * time.Millisecond)
+
+		// Get result second time - should fail since channel is closed
+		_, err = handle.GetResult()
+		require.Error(t, err, "second GetResult should fail")
+		assert.Contains(t, err.Error(), "workflow result channel is already closed")
+	})
```
maxdml: No need this test -- already tested somewhere else

dbos/workflows_test.go
```diff
@@ -4523,3 +4528,82 @@ func TestWorkflowIdentity(t *testing.T) {
...
+
+	t.Run("WorkflowPollingHandleNoTimeout", func(t *testing.T) {
+		// Test without timeout - should work normally
+		handle, err := RunWorkflow(dbosCtx, slowWorkflow, 1*time.Millisecond)
+		require.NoError(t, err, "failed to start workflow")
+
+		result, err := handle.GetResult()
+		require.NoError(t, err, "GetResult without timeout should succeed")
+		assert.Equal(t, "done", result)
+	})
```
maxdml: No need this test -- we exercise this path plenty already

dbos/workflow.go
```diff
@@ -162,12 +178,33 @@ type workflowHandle[R any] struct {
...
+	case <-timeoutChan:
+		return *new(R), fmt.Errorf("workflow result timeout after %v", options.timeout)
```
maxdml: Let's wrap in the cancellation cause (`context.Cause(err)`), so the end user can do `error.As` or `error.Is` and find a standard deadline exceeded error.

dbos/workflow.go
maxdml: For this path to kick in, you need to manually Launch and Shutdown the DBOSContext in a test. The sequence would be:

- Launch
- Run blocking workflow (so that the function doesn't return and its result never reaches `outcomeChan`)
- GetResult with no timeout
- Shutdown DBOSContext
- GetResult should result the error (which should be a context.Context [DeadlineExceeded](https://pkg.go.dev/context#pkg-variables) error)

maxdml: For this to work, the workflow must signal that it started, which you can achieve by using the test utilities "event" API. See an example [here](https://github.com/dbos-inc/dbos-transact-golang/blob/28d50f82d43affd776ca128b5b7df2c6915e0bae/dbos/workflows_test.go#L1445).

PhakornKiong: For this case, should it be `context.Canceled` instead?

maxdml: You mean for a test exercising `case <-h.dbosContext.Done():` ? You're right, it depends on how the context is cancelled (manually or through a deadline)

PR #175, opened 2025-10-27
Improve workflow/steps input/outputs serializer

maxdml:
This PR:
- Lift encoding outside of the system database
- Improve the gob decoder to reconstruct pointer types automatically
- Performs gob registration lazily before encode/decode
- Distinguish between nil values and empty strings stored in the DB (we use a wrapper over the value to gob encode/decode)

What we do not support:
- Workflow signatures with interface types (works if the user manually `gob.Register` the concrete implementation of the interface)
- Workflow signatures with nested pointers

### Tests
- Getting result from direct handle, polling handle, listing workflows and getting wf steps
- send/recv, setEvent/getEvent
- workflow recovery, workflow queues
- Test workflows with a "concrete" signature, and a workflow with a typeless signature (call the later with many concrete types)

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -1260,11 +1260,14 @@ func TestWorkflowRecovery(t *testing.T) {
 
 		// At least 5 of the 2nd steps should have errored due to execution race
 		// Check they are DBOSErrors with StepExecutionError wrapping a ConflictingIDError
-		secondStepErrorsMu.Lock()
-		errorsCopy := make([]error, len(secondStepErrors))
-		copy(errorsCopy, secondStepErrors)
-		secondStepErrorsMu.Unlock()
-		require.GreaterOrEqual(t, len(errorsCopy), 5, "expected at least 5 errors from second steps due to recovery race, got %d", len(errorsCopy))
+		var errorsCopy []error
+		require.Eventually(t, func() bool {
+			secondStepErrorsMu.Lock()
+			errorsCopy := make([]error, len(secondStepErrors))
+			copy(errorsCopy, secondStepErrors)
+			secondStepErrorsMu.Unlock()
+			return len(errorsCopy) >= 5
+		}, 10*time.Second, 100*time.Millisecond)
 		for _, err := range errorsCopy {
 			dbosErr, ok := err.(*DBOSError)
 			require.True(t, ok, "expected error to be of type *DBOSError, got %T", err)


dbos/recovery.go
```diff
@@ -18,13 +14,6 @@ func recoverPendingWorkflows(ctx *dbosContext, executorIDs []string) ([]Workflow
 	}
 
 	for _, workflow := range pendingWorkflows {
-		if inputStr, ok := workflow.Input.(string); ok {
-			if strings.Contains(inputStr, "Failed to decode") {
-				ctx.logger.Warn("Skipping workflow recovery due to input decoding failure", "workflow_id", workflow.ID, "name", workflow.Name)
-				continue
-			}
-		}
```
maxdml, line 26: Unused

dbos/queues_test.go
```diff
@@ -540,8 +540,9 @@ func TestQueueRecovery(t *testing.T) {
 			// Root workflow case
 			result, err := h.GetResult()
 			require.NoError(t, err, "failed to get result from recovered root workflow handle")
-			castedResult, ok := result.([]int)
-			require.True(t, ok, "expected result to be of type []int for root workflow, got %T", result)
+			// Need to re-marshal to []int because JSON serialized it to []interface{}
+			castedResult, err := convertJSONToType[[]int](result)
+			require.NoError(t, err, "failed to convert result to []int")
```
maxdml: Because recovery handles are typeless, we won't be able to resolve that for the user -- that's ok because end-users are not expected to manipulate recovery handles.

dbos/workflow.go
```diff
@@ -499,21 +513,30 @@ func RegisterWorkflow[P any, R any](ctx DBOSContext, fn Workflow[P, R], opts ...
...
-		typedInput, ok := input.(P)
-		if !ok {
-			return nil, newWorkflowUnexpectedInputType(fqn, fmt.Sprintf("%T", typedInput), fmt.Sprintf("%T", input))
+		workflowID, err := GetWorkflowID(ctx)
+		if err != nil {
+			return *new(R), newWorkflowExecutionError("", fmt.Errorf("getting workflow ID: %w", err))
+		}
+		// In async execution (e.g., recoverPendingWorkflows and queue runner), we don't know the input type
+		// Here we can capture it and add the casting to the logic
+		// Note that the input might come from the database.
```
maxdml: as a matter of fact, it actually always comes from the database

kraftp (commented):
What can JSON support? What can't JSON support? That needs to be more explicit and documented. I sense that there are weird non-obvious failure cases especially when interface info is lost and I'd like to know where they are. We may even make the interface more restrictive.

dbos/system_database.go
```diff
@@ -1266,8 +1238,8 @@ type recordChildWorkflowDBInput struct {
 
 func (s *sysDB) recordChildWorkflow(ctx context.Context, input recordChildWorkflowDBInput) error {
 	query := fmt.Sprintf(`INSERT INTO %s.operation_outputs
-            (workflow_uuid, function_id, function_name, child_workflow_id)
-            VALUES ($1, $2, $3, $4)`, pgx.Identifier{s.schema}.Sanitize())
+            (workflow_uuid, function_id, function_name, child_workflow_id, output)
+            VALUES ($1, $2, $3, $4, '')`, pgx.Identifier{s.schema}.Sanitize())
```
maxdml: Required because when we run GetWorkflowSteps as step, we'd try to gob-encode the `StepInfo.Output` and get `gob: cannot encode nil pointer of type *string inside interface`.

In general we cannot store NULL values in the database if we want being able to gob-decode them in nested properties like the above.

dbos/workflows_test.go
```diff
@@ -1307,7 +1134,9 @@ func TestWorkflowRecovery(t *testing.T) {
 			return fmt.Sprintf("completed-%d", index), nil
 		}, WithStepName(fmt.Sprintf("BlockingStep-%d", index)))
 		if err != nil {
+			secondStepErrorsMu.Lock()
 			secondStepErrors = append(secondStepErrors, err)
+			secondStepErrorsMu.Unlock()
```
maxdml: fix a race in the test

dbos/system_database.go
kraftp: But this is actually NULL, so it should be stored as NULL. Database NULL should represent a value that isn't there. Why can't we do that? And how is this a "nested" property? It's just a nonexistent output, no nesting.

maxdml: You are correct, there is a way around this. What we faced is that gob cannot encode nil values. For example if you do

```golang
    var nilPtr *string = nil
    toEncode := gobValue{Value: nilPtr}

    var buf bytes.Buffer
    enc := gob.NewEncoder(&buf)
    if err := enc.Encode(toEncode); err != nil {
        return fmt.Errorf("encode error: %w", err)
    }
```

This will fail with `gob: cannot encode nil pointer of type *string inside interface`

We can run into this situation if the output field from `StepInfo` is nil.

```golang
type StepInfo struct {
	Output          any
}
```

Because we run `GetWorkflowSteps` within `RunAsStep`, we store its output in the database, and thus encode it. Encoding will fail with the error above.

What I'll be pushing soon is an intermediate layer of decoding in an internal stepInfo type where the output is a `*string`, then restoring the output into StepInfo.Output, which is `any`. That way gob will have enough type information to store the nil pointer.

While this fixes this use case, in general it means that we cannot encode a workflow input which is like

```golang
var nilPtr *string = nil
input := WorkflowInput{
	AnyField:     nilPtr, // Store nil pointer in any field
}
```

And detecting these inputs at runtime is very complicated (needs recursive reflection)

maxdml: And btw this extends to storing composite types in `any` fields, e.g., `[]*int`. Albeit these could be manually registered, one by one.

All of these cases are known `encoding/gob` limitations.

kraftp:
> > I'm probably missing it, but where in this PR enforces what can and can't be serialized?
> 
> There is actually no way to declare a generic type that exclude all interfaces or nested pointers.
> 
> We'd have to do it at runtime, both in `RegisterWorkflow` (as early as it gets) and `RunAsStep` (late feedback.)
> 
> But doing it for interfaces would also prevent users from registering themselves the interface. Manual registration from the user makes the whole thing work. I think we should just clearly document it.
> 
> Do you think we should error / panic at runtime if we identify nested pointers?

It's fine if there's no compile-time way to identify unserializable types. There isn't any in the other languages either. But we should clearly document what can and can't be serialized (ideally supplemented with links to gob docs) and should throw clear errors if we get something that isn't serializable.

kraftp (commented):
Just to be clear, what's wrong with nested pointers?

maxdml:
> Just to be clear, what's wrong with nested pointers?

On the `GetWorkflowSteps` and `ListWorkflows`, we decode values into an `any` variable. gob cannot decode into `any` a value that was encoded from a concrete type. To address this, we wrap every encoded value in:

```golang
type gobValue struct {
    value any
}
```

Doing so allows Gob to decode anything into an `any` value (the get steps/ list wf paths).

On the typed paths (e.g., `RetrieveWorkflow[T]`), because we don't automatically decode into `T`, we must reconstruct it using reflection. If `T` is a pointer type, we must dynamically detect it and re-create a pointer type (we can't just do `value.(T)`). I elected to not perform recursive reflection on `T` to identify the nesting level (this is un-haltable.)

maxdml:
closing in favor of https://github.com/dbos-inc/dbos-transact-golang/pull/183

PR #183, opened 2025-11-10
Json serializer

maxdml:
- use JSON for encoding/decoding
- lift encode/decode out of the system database
- distinguish nil values (stored as NULL) in Postgres
- return json strings on the ListWorkflows GetWorkflowSteps paths
- Add comprehensive testing suite

On the get wf steps / list wf path, the contract is that if the value is nil, you know it was storing a nil value. Else, the value is a JSON string of your workflow's or step's `T`.

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -79,6 +79,10 @@ type workflowOutcome[R any] struct {
 	needsDecoding bool // true if result came from awaitWorkflowResult (ID conflict path) and needs decoding
 }
 
+type stepCheckpointedOutcome struct {
+	value any // The encoded value (should be a *string)
+}
+
 // WorkflowHandle provides methods to interact with a running or completed workflow.
 // The type parameter R represents the expected return type of the workflow.
 // Handles can be used to wait for workflow completion, check status, and retrieve results.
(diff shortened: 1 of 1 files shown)


dbos/client_test.go
```diff
@@ -369,9 +370,14 @@ func TestCancelResume(t *testing.T) {
 
 		// Wait for workflow completion
 		proceedSignal.Set() // Allow the workflow to proceed to step two
-		result, err := resumeHandle.GetResult()
+		resultAny, err := resumeHandle.GetResult()
```
maxdml, line 372: We could make the resume handle typed so the returned polling handle can be of type T as well. (Right now on this path, the user gets a polling handle typed any, and the result is already json decoded into an any value, losing its type information)

dbos/queues_test.go
```diff
@@ -538,10 +539,16 @@ func TestQueueRecovery(t *testing.T) {
...
-			castedResult, ok := result.([]int)
-			require.True(t, ok, "expected result to be of type []int for root workflow, got %T", result)
+			// re-encode and decode the result from []interface{} to []int
+			encodedResult, ok := resultAny.([]any)
+			require.True(t, ok, "expected result to be a []any")
+			jsonBytes, err := json.Marshal(encodedResult)
+			require.NoError(t, err, "failed to marshal result to JSON")
+			var castedResult []int
+			err = json.Unmarshal(jsonBytes, &castedResult)
+			require.NoError(t, err, "failed to decode result to []int")
```
maxdml, line 551: on this path the result was already decoded into any, which lost type info. This is because recovery handles are untyped. Just needed for testing as recovery handles are not public.

dbos/serialization.go
```diff
@@ -1,68 +1,116 @@
...
+func newJSONSerializer[T any]() serializer[T] {
+	return &jsonSerializer[T]{}
 }
 
-func deserialize(data *string) (any, error) {
-	if data == nil || *data == "" {
+func (j *jsonSerializer[T]) Encode(data T) (*string, error) {
+	// Check if the value is nil (for pointer types, slice, map, etc.)
+	if isNilValue(data) {
+		// For nil values, return nil pointer
```
maxdml, line 24: In Go, nil values are not necessarily pointers. We need the helper function for the case where T is `any`.

```golang
var p *int  // p is nil
var v any = p  // v is NOT nil (it's an interface containing a nil pointer)

fmt.Println(p == nil)  // true
fmt.Println(v == nil)  // false! The interface itself is not nil
```

dbos/system_database.go
```diff
@@ -1431,37 +1403,32 @@ func (s *sysDB) checkOperationExecution(ctx context.Context, input checkOperatio
...
-	Output          any    // The output returned by the step (if any)
-	Error           error  // The error returned by the step (if any)
-	ChildWorkflowID string // The ID of a child workflow spawned by this step (if applicable)
+type stepInfo struct {
+	StepID          int     // The sequential ID of the step within the workflow
+	StepName        string  // The name of the step function
+	Output          *string // The output returned by the step (if any)
+	Error           error   // The error returned by the step (if any)
+	ChildWorkflowID string  // The ID of a child workflow spawned by this step (if applicable)
 }
```
maxdml, line 1424: Internal struct so we can distinguish easily between empty outputs and nil outputs. In v1 we can modify `StepInfo`'s `Output` field to be of type `*string` instead of `any`.

dbos/workflow.go
```diff
@@ -205,11 +206,16 @@ func (h *workflowHandle[R]) GetResult(opts ...GetResultOption) (R, error) {
...
 	// If we are calling GetResult inside a workflow, record the result as a step result
 	workflowState, ok := h.dbosContext.Value(workflowStateKey).(*workflowState)
 	isWithinWorkflow := ok && workflowState != nil
 	if isWithinWorkflow {
-		encodedOutput, encErr := serialize(outcome.result)
+		if _, ok := h.dbosContext.(*dbosContext); !ok {
+			return *new(R), newWorkflowExecutionError(workflowState.workflowID, fmt.Errorf("invalid DBOSContext: expected *dbosContext"))
+		}
+		serializer := newJSONSerializer[R]()
+		encodedOutput, encErr := serializer.Encode(decodedResult)
```
maxdml, line 222: as an optimization we could pass the encoded value around the `workflowOutcome`...

dbos/workflow.go
```diff
@@ -1319,23 +1408,51 @@ func Recv[R any](ctx DBOSContext, topic string, timeout time.Duration) (R, error
 	if err != nil {
 		return *new(R), err
 	}
-	// Type check
+
+	// Handle nil message
+	if msg == nil {
+		return *new(R), nil
+	}
```
maxdml, line 1422: Technically should never happen because we have a not null constraint on the message column.

dbos/workflows_test.go
```diff
@@ -1818,27 +1678,34 @@ func TestSendRecv(t *testing.T) {
 	Launch(dbosCtx)
 
 	t.Run("SendRecvSuccess", func(t *testing.T) {
```
maxdml, line 1680: Fixed a race in this suite.

kraftp (approved):

PR #184, opened 2025-11-12
Distinguish nil values in the database

maxdml:
Use a special marker `__DBOS_NIL` to represent nil values in the database instead of storing as `NULL`.


kraftp (commented):
Is this necessary? What problem is it solving? I thought we decided not to do this. The problem it could solve is UI confusion between nil and empty, but I don't see the code for that here and I'm not even sure it's a problem in Go because Go returns (nil, err) and such.

kraftp (approved):

PR #193, opened 2025-11-21
make awaitWorkflowResult polling interval configurable

shirzady1934:
This PR introduces a configurable polling interval for `awaitWorkflowResult`, enabling users to control how frequently `GetResult()` polls the database for workflow completion. Previously, the polling interval was fixed at `_DB_RETRY_INTERVAL`, with no way to adjust it. This change adds flexibility for workloads that require faster responsiveness or reduced DB query pressure.

###  Functional Option: `WithPollingInterval()`
- Adds a new functional option for `GetResult()`:
  ```go
  func WithPollingInterval(interval time.Duration) GetResultOption

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -120,9 +120,9 @@ func WithHandleTimeout(timeout time.Duration) GetResultOption {
 	}
 }
 
-// WithPollingInterval sets the polling interval for awaiting workflow completion in GetResult.
+// WithHandlePollingInterval sets the polling interval for awaiting workflow completion in GetResult.
 // If a non-positive interval is provided, the default interval is used.
-func WithPollingInterval(interval time.Duration) GetResultOption {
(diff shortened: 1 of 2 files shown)


maxdml:
> Thanks for the feedback, @maxdwell. I updated the PR to match your suggestion. I removed the separate `workflow_getresult_options_test.go` file and folded the polling-interval checks into the existing tests. The previous `TestWorkflowHandleTimeout` is now a small suite called `TestWorkflowHandles`, and both subtests use `WithPollingInterval(1 * time.Millisecond)` in their `GetResult` calls. This exercises the new option through normal workflow execution without inspecting internal fields or introducing new mocks, keeping the tests simple and consistent with your guidance. Let me know if you’d like any further adjustments.

thanks @shirzady1934 , lgtm. Two small things:
- Let's name the functional option `WithHandlePollingInterval` (we'll need a similar thing for queues)
- Please run go fmt on the test file

PR #194, opened 2025-11-23
Ensure executorID is not overridden if already set

YakirOren:
I've added a check to prevent overriding an existing executorID value.


maxdml:
@YakirOren Thanks but actually we don't want to enable manually setting an executor ID when using conductor, at this time. The reason is that this will invariably cause weird bugs with how conductor manages executors life cycle.

PR #195, opened 2025-11-27
Add queue option

tuannguyensn2001:
# Description
Currently queue interval is fixed value. This PR add option to custom these values  when initialize DbOS


maxdml:
> # Description
> Currently queue interval is fixed value. This PR add option to custom these values when initialize DbOS

Thanks @tuannguyensn2001. Let's pass the config through functional options please (because these should all be optional configurations)

Issue #203, opened 2025-12-08
Provide Metrics and Executor Info to Conductor


kraftp:
Analogous to https://github.com/dbos-inc/dbos-transact-ts/pull/1134

PR #204, opened 2025-12-08
Add workflow / step count metrics

maxdml:
address https://github.com/dbos-inc/dbos-transact-golang/issues/203

### DIFF
--- a/dbos/metrics_test.go
+++ b/dbos/metrics_test.go
@@ -1,0 +1,130 @@
+package dbos
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


kraftp:
> > Let's also add this to the Conductor-side tests (https://github.com/dbos-inc/dbos-conductor/blob/927ae5b5234f5759c228b8c9341798d3333cf948/tests/test_conductor.py#L506)
> 
> The conductor test already runs with a Go app. Am I missing something?
> Screenshot 2025-12-15 at 12 24 07

Look at the code--the test is skipped

maxdml:
Note that the conductor PR will need this one to be merged first (conductor checks out the latest `main` branch of the Go SDK)

Issue #202, opened 2025-12-08
Copy Events During Fork


kraftp:
Analogous to https://github.com/dbos-inc/dbos-transact-py/pull/511

PR #207, opened 2025-12-16
Copy events during fork

maxdml:
#202

### DIFF
--- a/dbos/dbos_test.go
+++ b/dbos/dbos_test.go
@@ -251,7 +251,7 @@ func TestConfig(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, exists, "dbos_migrations table should exist")
 
-		// Verify migration version is 5 (after initial migration, queue partition key migration, workflow status index migration, forked_from migration, and step timestamps migration)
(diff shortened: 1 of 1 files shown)


dbos/system_database.go
```diff
@@ -2091,6 +2124,17 @@ func (s *sysDB) setEvent(ctx context.Context, input WorkflowSetEventInput) error
 		return fmt.Errorf("failed to insert/update workflow event: %w", err)
 	}
 
+	// Record event in workflow_events_history
+	insertHistoryQuery := fmt.Sprintf(`INSERT INTO %s.workflow_events_history (workflow_uuid, function_id, key, value)
+					VALUES ($1, $2, $3, $4)
+					ON CONFLICT (workflow_uuid, function_id, key)
+					DO UPDATE SET value = EXCLUDED.value`, pgx.Identifier{s.schema}.Sanitize())
```
maxdml, line 2134: Unclear to me when we'd be expecting a conflict on this path.

qianl15: I think it was just a copy-pasted thing from above. Shouldn't really conflict as you said.

Issue #205, opened 2025-12-09
Implement Patching


kraftp:
Analogous to https://github.com/dbos-inc/dbos-transact-py/pull/524

PR #210, opened 2025-12-18
Patch

maxdml:
#205 

Allow running DBOS in patching mode.

In this mode, the application version is set to PATCHING_ENABLED, by default.

To use patching, use the new methods Patch() and DeprecatePatch(). See the new tests in workflows_test.go for usage.

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -78,8 +78,8 @@ func processConfig(inputConfig *Config) (*Config, error) {
 		dbosConfig.DatabaseSchema = _DEFAULT_SYSTEM_DB_SCHEMA
 	}
 
-	// If patching is enabled, fix the application version
-	if dbosConfig.EnablePatching {
+	// If patching is enabled and application version is not set, fix the application version
+	if dbosConfig.EnablePatching && dbosConfig.ApplicationVersion == "" {
 		dbosConfig.ApplicationVersion = "PATCHING_ENABLED"
 	}
 
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -4883,4 +4883,55 @@ func TestPatching(t *testing.T) {
 		require.Contains(t, dbosErr2.Message, "Patching system is not enabled", "expected error message to mention patching is not enabled")
 		require.Contains(t, dbosErr2.Message, "EnablePatching", "expected error message to mention EnablePatching")
 	})
+
+	t.Run("PatchingEnabledWithVersioning", func(t *testing.T) {
+		databaseURL := getDatabaseURL()
+
+		t.Run("PreservesApplicationVersionWhenSetInConfig", func(t *testing.T) {
+			// Clear env vars to ensure we're testing config values
+			t.Setenv("DBOS__APPVERSION", "")
+			resetTestDatabase(t, databaseURL)
+
+			// Create a DBOS context with patching enabled and a custom application version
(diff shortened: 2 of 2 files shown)


dbos/workflow.go
```diff
@@ -1646,6 +1647,111 @@ func Sleep(ctx DBOSContext, duration time.Duration) (time.Duration, error) {
...
 
+func (c *dbosContext) Patch(_ DBOSContext, patchName string) (bool, error) {
+	if !c.config.EnablePatching {
+		return false, newPatchingNotEnabledError()
+	}
+
+	if patchName == "" {
+		return false, errors.New("patch name cannot be empty")
+	}
+
```
kraftp, line 1660: Should prefix the patch name (for example with `DBOS.patch`) to make workflows easier to visualize

dbos/workflow.go
```diff
@@ -1646,6 +1647,111 @@ func Sleep(ctx DBOSContext, duration time.Duration) (time.Duration, error) {
...
+			stepID:     wfState.stepID + 1,
+			patchName:  patchName,
+		})
+	}, withRetrierLogger(c.logger))
+
+	if patchNameFromDB != patchName || err == pgx.ErrNoRows {
+		return true, nil
+	}
+	wfState.nextStepID()
+	return false, err
```
kraftp: What does it mean to return false here?

dbos/workflows_test.go
```diff
@@ -4610,3 +4610,242 @@ func TestWorkflowHandleContextCancel(t *testing.T) {
...
+			OriginalWorkflowID: patchedHandle.GetWorkflowID(),
+			StartStep:          3,
+		})
+		require.NoError(t, err, "failed to fork workflow")
+		result, err = forkHandle.GetResult()
+		require.NoError(t, err, "failed to get result")
+		require.Equal(t, 3, result, "expected result to be 3")
+		steps, err = GetWorkflowSteps(dbosCtx, forkHandle.GetWorkflowID())
+		require.NoError(t, err, "failed to get workflow steps")
+		require.Equal(t, 4, len(steps), "expected 4 steps")
```
kraftp: Also test that forking an old workflow with the original code version fails cleanly

dbos/workflow.go
maxdml: Ha good point, we don't need this function to return a bool, at all. It can just return an error.

dbos/workflows_test.go
maxdml: Just to clarify, you want to exercise the path where the code has been patched, but the user wants to fork a workflow that ran the previous code version, and only the new code version is present in the workflow registry? Which would result in a non-determinism error when the new step name differs. (This error will happen only if the new code version ends up having different step names.)

kraftp: Yes, I want to see a clean non-determinism error, like in the Python tests.

maxdml: OK, but note this has not much to do with patching, but rather with forking from the wrong version. This could happen when using application versions.

dbos/dbos.go
```diff
@@ -76,6 +78,11 @@ func processConfig(inputConfig *Config) (*Config, error) {
 		dbosConfig.DatabaseSchema = _DEFAULT_SYSTEM_DB_SCHEMA
 	}
 
+	// If patching is enabled, fix the application version
+	if dbosConfig.EnablePatching {
```
kraftp: If patching is enabled and the application version is not being set

Issue #208, opened 2025-12-16
Support opting-out of listen/notify in the system db


maxdml:
Some durability backends (like cockroach) don't support (well) listen/notify. Our current event and notification systems use these. This issue is about allowing users to opt-out (should be on by default) of listen/notify for `recv` and `getEvent`, and use polling instead.

PR #214, opened 2025-12-19
Support for CockroachDB

maxdml:
This PR:
- Adds ability to detect whether we are running cockroachdb and condition behavior if so (see bellow)
- Runs the entire test suite with a password-less cockroachdb container (single node)
- Adds missing retry/retryWithResult on functions that call the DB.

CockroachDB behavior:
- No WITH FORCE
- No registration of DB triggers
- Use a poller to retrieve notifications/events
- Requires direct conversion before arithmitic in SQL statements

#208

### DIFF
--- a/cmd/dbos/cli_integration_test.go
+++ b/cmd/dbos/cli_integration_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -40,7 +41,6 @@ const (
 )
 
 // getDatabaseConfig extracts database connection parameters from standard PostgreSQL environment variables
-// Uses PGHOST, PGPORT, PGDATABASE, PGUSER, PGPASSWORD, PGSSLMODE (see https://www.postgresql.org/docs/current/libpq-envars.html)
 // Always defaults to sslmode=disable for test setups
 func getDatabaseConfig(dbRole string) (user, password, host, port, dbName, sslmode string) {
 	// Read from standard PostgreSQL environment variables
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-crdb.yml
maxdml, line 1: Run the entire test suite with a single, passwordless cockroachDB container

dbos/system_database.go
```diff
@@ -327,11 +336,11 @@ func newSystemDatabase(ctx context.Context, inputs newSystemDatabaseInput) (syst
 		if err != nil {
 			return nil, fmt.Errorf("failed to validate custom pool: %v", err)
 		}
+		defer poolConn.Release()
 		err = poolConn.Ping(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to validate custom pool: %v", err)
 		}
-		poolConn.Release()
 		pool = customPool
```
maxdml, line 344: Safer: if Ping fails we'll never release this connection.

dbos/system_database.go
```diff
@@ -1909,6 +1944,109 @@ func (s *sysDB) notificationListenerLoop(ctx context.Context) {
 	}
 }
 
+func (s *sysDB) notificationPollerLoop(ctx context.Context) {
```
maxdml, line 1947: This whole thing can be optimized if we ever need it:
- backoff policy
- batch reads

dbos/migrations/1_initial_dbos_schema.sql
```diff
@@ -1,6 +1,3 @@
--- Enable uuid extension for generating UUIDs
-CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
-
```
maxdml, line 3: unused

Issue #90, opened 2025-08-29
Support calling runAsStep in goroutines inside workflows


maxdml:
Because goroutines execution is non deterministic, to run workflow steps inside goroutines, we need to make the `StepID` generation deterministic. This could be done with a "durable `wg.Go()`", for example.

rishi-jat:
@maxdml I want to work on this issue.

af-md:
I would also like to work on this

I am happy to collaborate with @rishi-jat :)

Issue #89, opened 2025-08-29
durable select


maxdml:
Currently, using `select` in workflows is prohibited because `select` is non deterministic. A durable select would be a special type of step that records its outcome (the `case` that was selected)

PR #221, opened 2026-01-14
Allow running steps concurrently

maxdml:
Addresses #90.

This PR allows running steps inside goroutines. Because goroutines are scheduled non-deterministically, using the `go` keyword naïvely, e.g.:

```go
for range := i {
    go dbos.RunAsStep(...)
}
```

Can result in non-deterministic re-execution, that is, it is not guaranteed that the steps will start executing, with the same step ID, in the same order every time the workflow executes.

This PR adds a new `dbos.Go()` primitive that pre-generates a step ID before calling the `go` keyword. `dbos.Go()` returns a buffered channel of size 1 that the user can use to obtain the step result.

```go
resultChan, err := dbos.Go(ctx, func(ctx context.Context) (string, error) {
    return "Hello, World!", nil
})

resultChan := <-resultChan // wait for the channel to receive
if resultChan.err != nil {
    // Handle error
}
```

Note that using these channels with a select statement still requires the implementation of #89 for deterministic execution.

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1510,36 +1510,11 @@ func Go[R any](ctx DBOSContext, fn Step[R], opts ...StepOption) (chan StepOutcom
 			return
 		}
 
-		var typedResult R
-		// First check if this is a checkpointed outcome (encoded value from database)
-		if checkpointed, ok := outcome.Result.(stepCheckpointedOutcome); ok {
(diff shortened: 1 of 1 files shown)


kraftp (approved):
This looks correct to me, as an elegant way to get step-wise concurrency in your workflows. I hope the interface will be natural and familiar for Go devs.

PR #223, opened 2026-01-15
Durable select

maxdml:
Allow polling on a list of channels obtained from dbos.Go()

```golang
ch1, _ := dbos.Go(ctx, func(ctx context.Context) (string, error) { return "result1", nil })
ch2, _ := dbos.Go(ctx, func(ctx context.Context) (string, error) { return "result2", nil })
outcome, err := dbos.Select(ctx, []<-chan dbos.StepOutcome[string]{ch1, ch2})
```

The channels must all be of the same type `T`

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1583,17 +1583,34 @@ func Select[R any](ctx DBOSContext, channels []<-chan StepOutcome[R]) (R, error)
 	}
 
 	// Convert typed channels to any channels for internal processing
+	// Create a context that will be cancelled when Select completes to prevent goroutine leaks
+	selectCtx, cancelSelect := context.WithCancel(ctx)
+	defer cancelSelect() // Ensure all goroutines are cancelled when Select completes
+
 	anyChannels := make([]<-chan StepOutcome[any], len(channels))
 	for i := range channels {
 		anyCh := make(chan StepOutcome[any], cap(channels[i]))
-		// Use a closure with proper variable capture
 		srcCh := channels[i]
 		go func() {
 			defer close(anyCh)
-			for outcome := range srcCh {
-				anyCh <- StepOutcome[any]{
-					Result: outcome.Result,
-					Err:    outcome.Err,
+			for {
+				select {
+				case <-selectCtx.Done():
+					return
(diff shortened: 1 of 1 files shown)


dbos/workflow.go
```diff
@@ -1545,14 +1545,193 @@ func (c *dbosContext) Go(ctx DBOSContext, fn StepFunc, opts ...StepOption) (chan
...
+			defer close(anyCh)
+			for outcome := range srcCh {
+				anyCh <- StepOutcome[any]{
+					Result: outcome.Result,
+					Err:    outcome.Err,
+				}
+			}
+		}()
+		anyChannels[i] = anyCh
+	}
```
maxdml, line 1618: We need to do this to offload the work to an `any` `DBOSContext.Select`. This does not change the behavior of select (pick a ready channel uniformly at random over the list of provided channels), but adds more goroutine. Go being designed for a large number of lightweight goroutines, I'm not too worried about that. Note we cancel all the goroutines once the select returns, so we don't leak goroutines.

dbos/dbos.go
```diff
@@ -258,9 +259,34 @@ func WithoutCancel(ctx DBOSContext) DBOSContext {
...
+			executorID:              dbosCtx.executorID,
+			applicationID:           dbosCtx.applicationID,
+			queueRunner:             dbosCtx.queueRunner,
+		}
+		childCtx.launched.Store(launched)
+		return childCtx, cancelCauseFunc
+	}
+	return nil, func(error) {}
+}
+
```
maxdml, line 289: similar to https://pkg.go.dev/context#WithCancelCause, for `DBOSContext`

dbos/workflow.go
```diff
@@ -1208,8 +1208,8 @@ func WithNextStepID(stepID int) StepOption {
 // StepOutcome holds the result and error from a step execution
 // This struct is returned as part of a channel from the Go function when running the step inside a Go routine
 type StepOutcome[R any] struct {
-	result R
-	err    error
+	Result R     `json:"result"`
+	Err    error `json:"err"`
```
maxdml, line 1212: This is not necessary but, because the struct itself is exported, it could be useful to actually export the fields

Issue #216, opened 2026-01-12
Recv does not return an error on timeout


alejodiazg:
According to
https://docs.dbos.dev/golang/tutorials/workflow-communication#recv

Recv should be returning an error when provided timeout is reached.

If we go to the underlying implementation
https://github.com/dbos-inc/dbos-transact-golang/blob/main/dbos/system_database.go#L2014
no error is set, so we return nil error and a nil message string

which is the correct expected behavior, should the documentation be updated or the  implementation is incorrect?

maxdml:
Thanks for pointing this out. The implementation is correct and the documentation should be updated.

alejodiazg:
I would say its closer to whats expected idiomatically from go, an explicit error.
And also would respect the original intent of the documentation and comments.
In this case its also confusing as the step is marked as completed right?, so theres no status difference between the "Recieved in the time allocated" and "I timed out waiting"

maxdml:
Agreed it'd be more go idiomatic (return clear errors) and allow to differentiate "empty strings" and "it timed out". I'll take it on this week.

PR #224, opened 2026-01-16
Timeout errors

maxdml:
Address #216 

This is a breaking change: `dbos.Recv` and `dbos.GetEvent` will now return an error if they hit a timeout.

Also add handling potential loss of PG notification if PG goes down.

### DIFF
--- a/dbos/conductor.go
+++ b/dbos/conductor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"log/slog"
 	"math"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


chaos_tests/chaos_test.go
maxdml, line 1: use an event instead of sleeping

dbos/dbos_test.go
maxdml, line 1: use an event instead of sleeping

dbos/system_database.go
```diff
@@ -2052,14 +2082,24 @@ func (s *sysDB) recv(ctx context.Context, input recvInput) (*string, error) {
...
 		case <-done:
+			break loop
 		case <-time.After(timeout):
+			timeoutOccurred = true
 			s.logger.Warn("Recv() timeout reached", "payload", payload, "timeout", input.Timeout)
+			break loop
+		case <-repollChannel:
+			s.logger.Warn("Receive polling after repoll channel signal", "payload", payload)
+			// We were instructed to poll again because the connection was disconnected
+			err = s.pool.QueryRow(ctx, query, destinationID, topic).Scan(&exists)
```
chuck-dbos, line 2093: Should we have a polling occasionally even if we aren't told to poll?  (This mode is the mode used for CockroachDB.)

maxdml: For CockroachDB, I was going to go with the same approach we did in Python: have a notifier loop based on polling rather than listen/notify (here is the draft PR: https://github.com/dbos-inc/dbos-transact-golang/pull/214)

chuck-dbos: I see.

chuck-dbos (approved):

PR #228, opened 2026-01-19
Durable streams

maxdml:
Durable streams.

Writing to a stream can happen within a workflow (exactly once) or within a step (at least once).

Reading to a stream is not durable. It is a blocking operation that'll exit whenever the stream is closed or the writing workflow exits.

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -5362,10 +5362,7 @@ func writeStreamWorkflow(ctx DBOSContext, input struct {
 
 	// Write from step level with custom step name
 	_, err := RunAsStep(ctx, func(stepCtx context.Context) (string, error) {
-		// Note we don't expect users to be able to do this, because dbosContext is private.
-		// It is also pretty clear that WriteStream is a special step.
-		// This "boon" comes from the delicateness of _having_ to pass the step context to the step function (_not_ the workflow context)
-		return "", WriteStream(stepCtx.(*dbosContext), input.StreamKey, "step-value")
+		return "", WriteStream(stepCtx.(DBOSContext), input.StreamKey, "step-value")
 	}, WithStepName("not-just-write"))
 	if err != nil {
 		return "", err


dbos/client.go
```diff
@@ -322,6 +323,59 @@ func (c *client) GetWorkflowSteps(workflowID string) ([]StepInfo, error) {
...
+// Example:
+//
+//	values, closed, err := dbos.ClientReadStream[string](client, "workflow-id", "my-stream")
+//	if err != nil {
+//	    return err
+//	}
+//	for _, value := range values {
+//	    log.Printf("Stream value: %s", value)
+//	}
+func ClientReadStream[R any](c Client, workflowID string, key string) ([]R, bool, error) {
```
maxdml, line 352: The name `ReadStream` is already used at the package level.

With serialization moving to be custom, I think we should remove the client entirely and ask users to use a normal DBOS Context that they won't launch. The main idea behind having a client was to not having to explain two things:
- Context must not be launched
- Input/outputs must not be loaded

Custom serialization should remove the 2nd constraint, making the whole thing much easier to explain.

maxdml:
> This doesn't seem to support writing to a stream from a step as in the other languages--is there a reason for the change?

You can write to a stream from a step, it'll just be considered as part of the step (like we do for other occurence of "steps within steps")

https://github.com/dbos-inc/dbos-transact-golang/pull/228/changes#diff-4e36655cba483f14f99005ce843b724306a0e0a260243d976f131929bf773791R5364

kraftp:
> > This doesn't seem to support writing to a stream from a step as in the other languages--is there a reason for the change?
> 
> You can write to a stream from a step, it'll just be considered as part of the step (like we do for other occurence of "steps within steps")
> 
> https://github.com/dbos-inc/dbos-transact-golang/pull/228/changes#diff-4e36655cba483f14f99005ce843b724306a0e0a260243d976f131929bf773791R5364

I saw that, but that's not really a public interface and the syntax is horrifying

maxdml:
> > > This doesn't seem to support writing to a stream from a step as in the other languages--is there a reason for the change?
> > 
> > 
> > You can write to a stream from a step, it'll just be considered as part of the step (like we do for other occurence of "steps within steps")
> > https://github.com/dbos-inc/dbos-transact-golang/pull/228/changes#diff-4e36655cba483f14f99005ce843b724306a0e0a260243d976f131929bf773791R5364
> 
> I saw that, but that's not really a public interface and the syntax is horrifying

I'm not sure I understand the issue. There is a single write stream function that you can call from a workflow or from a step. In the former case it'll become a step automatically, in the latter case it'll be part of the calling step. What else do we need?

kraftp (approved):

PR #230, opened 2026-01-20
Debouncer

maxdml:
Implementes debouncing. Few notes:
- A debouncer is generic and must be created before launching DBOS. (Generic allows returning a typed workflow handle)
- On the client path, we must run the debouncer workflow as `any,any`, because the server has no way to know what concrete type the debouncing workflow is

Also add:
- Add ability to filter workflows per deduplication ID in `ListWorkflows`
- Some error logging

Also fixes:
- Client needs to `launch` the system DB to start its notification listener

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -2038,13 +2038,13 @@ func receiveIdempotencyWorkflow(ctx DBOSContext, topic string) (string, error) {
 func durableRecvSleepWorkflow(ctx DBOSContext, topic string) (string, error) {
 	// First Recv with 2-second timeout (will timeout)
 	msg1, err := Recv[string](ctx, topic, 2*time.Second)
-	if err != nil && !errors.Is(err, &DBOSError{Code: TimeoutError}) {
+	if err != nil && !strings.Contains(err.Error(), fmt.Sprintf("DBOS Error %d", TimeoutError)) {
 		return "", fmt.Errorf("unexpected error in first recv: %w", err)
 	}
 
 	// Second Recv with 2-second timeout (will also timeout)
 	msg2, err := Recv[string](ctx, topic, 2*time.Second)
-	if err != nil && !errors.Is(err, &DBOSError{Code: TimeoutError}) {
+	if err != nil && !strings.Contains(err.Error(), fmt.Sprintf("DBOS Error %d", TimeoutError)) {
 		return "", fmt.Errorf("unexpected error in second recv: %w", err)
 	}
 
@@ -2495,6 +2495,8 @@ func TestSendRecv(t *testing.T) {
 		require.Equal(t, 0, steps[0].StepID, "expected first step ID to be 0")
 		require.Equal(t, "DBOS.recv", steps[0].StepName, "expected first step to be recv")
 		require.Nil(t, steps[0].Output, "expected first recv step output to be nil (timeout)")
+		require.NotNil(t, steps[0].Error, "expected first recv step to have an error")
+		require.Contains(t, steps[0].Error.Error(), "DBOS.recv timed out", "expected step 0 to contain 'DBOS.recv timed out' in error message")
 		require.False(t, steps[0].StartedAt.IsZero(), "expected recv step to have StartedAt set")
 		require.False(t, steps[0].CompletedAt.IsZero(), "expected recv step to have CompletedAt set")
 		require.True(t, steps[0].CompletedAt.After(steps[0].StartedAt) || steps[0].CompletedAt.Equal(steps[0].StartedAt),
@@ -2512,6 +2514,8 @@ func TestSendRecv(t *testing.T) {
 		require.Equal(t, 2, steps[2].StepID, "expected third step ID to be 2")
 		require.Equal(t, "DBOS.recv", steps[2].StepName, "expected third step to be recv")
 		require.Nil(t, steps[2].Output, "expected second recv step output to be nil (timeout)")
+		require.NotNil(t, steps[2].Error, "expected second recv step to have an error")
+		require.Contains(t, steps[2].Error.Error(), "DBOS.recv timed out", "expected step 2 to contain 'DBOS.recv timed out' in error message")
 		require.False(t, steps[2].StartedAt.IsZero(), "expected recv step to have StartedAt set")
 		require.False(t, steps[2].CompletedAt.IsZero(), "expected recv step to have CompletedAt set")
 		require.True(t, steps[2].CompletedAt.After(steps[2].StartedAt) || steps[2].CompletedAt.Equal(steps[2].StartedAt),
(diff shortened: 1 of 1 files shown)


dbos/client_test.go
```diff
@@ -1227,3 +1227,269 @@ func TestClientReadStream(t *testing.T) {
...
+		assert.Equal(t, handle1.GetWorkflowID(), handle4.GetWorkflowID(), "all handles should refer to the same workflow ID")
+		assert.Equal(t, handle1.GetWorkflowID(), handle5.GetWorkflowID(), "all handles should refer to the same workflow ID")
+
+		result, err := handle5.GetResult()
+		require.NoError(t, err, "failed to get result")
+		assert.Equal(t, "input-5", result, "result should match latest input")
+
+		// Verify execution happened at least 200ms after first call (200ms debounce in rapid succession)
+		elapsed := time.Since(startTime)
+		assert.GreaterOrEqual(t, elapsed, 200*time.Millisecond, "execution should take at least 200ms")
```
maxdml, line 1316: hard to say exactly how much time should elapse, so erring on the conservative side to avoid test flakiness

dbos/dbos.go
```diff
@@ -435,6 +435,11 @@ func NewDBOSContext(ctx context.Context, inputConfig Config) (DBOSContext, error
 
 	// Initialize the queue runner and register DBOS internal queue
 	initExecutor.queueRunner = newQueueRunner(initExecutor.logger)
+	NewWorkflowQueue(initExecutor, _DBOS_INTERNAL_QUEUE_NAME)
```
maxdml, line 438: This doesn't start any background resources, so OK to add it in the queue registry before launching.

dbos/debouncer.go
```diff
@@ -0,0 +1,511 @@
...
+const _DEBOUNCER_TOPIC = "_dbos_debouncer_topic"
+
+// debouncerInput is the input to the internal debouncer workflow
+type debouncerInput[P any] struct {
+	InitialInput                  P
+	TargetWorkflowFQNOrCustomName string
+	TargetWorkflowID              string
+	Delay                         time.Duration   // Time by which to delay workflow execution
+	Timeout                       time.Duration   // Maximum time before starting the workflow
+	WorkflowOptions               workflowOptions // Options to pass to target workflow (serializable)
```
maxdml, line 22: functions are not JSON serializable, including wf options, which we will resolve & pass directly to the debouncing workflow.

dbos/debouncer.go
```diff
@@ -0,0 +1,511 @@
...
+					Input: input,
+					Delay: delay,
+					ID:    messageID,
+				}, _DEBOUNCER_TOPIC)
+				if err != nil {
+					return nil, err
+				}
+
+				// Acknowledge the send by getting an event with the message ID
+				_, err = GetEvent[bool](ctx, debouncerWorkflowID, messageID, 2*time.Second) // XXX unclear what's a good timeout here.
```
maxdml: This 2 seconds is really arbitrary, and will likely be flaky at very high load. If we miss an event and the debouncing workflow moves on, potentially starting the user workflow. We'll retry and start a new execution of the new workflow that, nevertheless, respects the delay.

dbos/debouncer.go
```diff
@@ -0,0 +1,511 @@
...
+
+	// We use the wrapped, type-erased workflow wrapper from the workflow registry that calls ctx.RunWorkflow
+	// Which doesn't do any pre-encoding of the input, and calls a type-erased function that expects an encoded input
+	// So we need to serialize the input here
+	workflowOpts = append(workflowOpts, withAlreadyEncodedInput())
+	serializer := newJSONSerializer[P]()
+	encodedInput, err := serializer.Encode(currentInput)
+	if err != nil {
+		return zero, fmt.Errorf("failed to serialize input: %w", err)
+	}
```
maxdml, line 496: So far the registry was always called with inputs straight from the DB, encoded (queue runner, recovery). This changes so we 1) have to encode ahead of time to use the wrapped workflow function and 2) introduce `withAlreadyEncodedInput` to avoid double encoding during `RunWorkflow`.

dbos/workflow.go
```diff
@@ -621,17 +621,18 @@ type Workflow[P any, R any] func(ctx DBOSContext, input P) (R, error)
...
+	WorkflowID          string
+	QueueName           string
+	ApplicationVersion  string
+	MaxRetries          int
+	DeduplicationID     string
+	Priority            uint
+	AuthenticatedUser   string
+	AssumedRole         string
+	AuthenticatedRoles  []string
+	QueuePartitionKey   string
```
maxdml, line 634: Exported (but the struct is still private) to enable JSON serialization.

dbos/workflow.go
```diff
@@ -679,35 +680,44 @@ func WithPriority(priority uint) WorkflowOption {
...
 	}
 }
 
 // An internal option we use to map the reflection function name to the registration options.
 func withWorkflowName(name string) WorkflowOption {
 	return func(p *workflowOptions) {
-		p.workflowName = name
+		if p.WorkflowName == "" {
+			p.WorkflowName = name
+		}
```
maxdml, line 692: Allows not overriding previous calls to `withWorkflowName`

dbos/debouncer.go
```diff
@@ -0,0 +1,495 @@
...
+			}
+			return newWorkflowPollingHandle[R](dc.Client.(*client).dbosCtx, decodedInput.TargetWorkflowID), nil
+		}
+		return nil, err
+	}
+}
+
+// internalDebouncerWF is the internal workflow that implements debouncing logic.
+// It collects inputs, delays execution, and runs the target workflow with the latest input.
+func internalDebouncerWF[P any, R any](ctx DBOSContext, input debouncerInput[P]) (R, error) {
```
kraftp, line 363: This is a workflow, so it should be deterministic. If the tests are passing with this implementation, we also may need more tests around recovery.

maxdml: Ha, I assume you are referring about the `time.Now()` that's not done in a step? Fixing it now.

With respect to recovery tests I think the easiest way to test this would be chaos tests: we can't inject blocks/failures in this workflow code very easily (as we do with user provided workflows).

kraftp: Perhaps, but you can still verify determinism by doing what the Python test does [here](https://github.com/dbos-inc/dbos-transact-py/blob/c1ba8e30fd364e4eed99990cd0848ca3f3cdf085/tests/test_debouncer.py#L69), running the test workflow again from the beginning.

maxdml: I see, if remaining time is botched because of non determinism, some steps (Recv, SetEvent) might be missing. I'll add that. And that's easy to do because we can replay a workflow that's parent to the internal debouncing workflow. Neat!

maxdml: Interesting. Two things I'm noticing:
- The python test doesn't replay the debounce workflow. It replays the parent workflow that gets a checkpointed result for the debounce workflow.
- Even if it did, our checking for operation outputs only checks if the function name is the expected one for a given step ID. In this case, the step names / IDs would be the same, but the step input wouldn't be. Also, if we replay a workflow and executed _less_ steps, we wouldn't be able to tell as well, because we only check when recording a step output.

That's why this case is so tricky to test: not having the sleep as a step can only result in the Recv timeout parameter being different, or the Recv not running at all.

PR #236, opened 2026-01-24
Read stream async

### LINKED ISSUE #234: Real-time stream events from a workflow
I'm experimenting with the new Streams in https://github.com/dbos-inc/dbos-transact-golang/pull/228

The [TS docs](https://docs.dbos.dev/typescript/tutorials/workflow-communication#workflow-streaming) say:

> Workflows can stream data in real time to clients. This is useful for streaming results from a long-running workflow or LLM call or for monitoring or progress reporting.

And the TS `readStream` method is an AsyncGenerator.

But the Go implementation says:

```
ReadStream(_ DBOSContext, workflowID string, key string) ([]any, bool, error)                               // Read values from a durable stream (blocks until workflow inactive or stream closed)
```

Will there be a way to get individual stream events in real time in Go?

maxdml:
Provide async APIs to read "real time" from a stream. The underlying read stream logic now writes read values to a channel, that is either fully drained (sync mode) or returned to the user (async mode).

fixes #234

### DIFF
--- a/dbos/client.go
+++ b/dbos/client.go
@@ -340,9 +340,8 @@ func (c *client) ClientReadStream(workflowID string, key string) ([]any, bool, e
 }
 
 // ClientReadStream reads values from a durable stream with type safety.
-// This method blocks until one of the following conditions is met:
-//   - The workflow becomes inactive (status is not PENDING or ENQUEUED)
-//   - The stream is closed (sentinel value is found)
+// This method blocks until the stream is closed or an error occurs.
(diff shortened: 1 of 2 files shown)


dbos/workflows_test.go
maxdml, line 1: changes here are about running the existing test suite in async mode.

dbos/client_test.go
maxdml, line 1: changes here are about running the existing test suite in async mode.

PR #238, opened 2026-01-28
Improve tests

maxdml:
Fix:
- Some of communication tests are flaky: most of the time they incur a sleep step, but this is not guaranteed
--> Make the producers run before the consumers so we guarantee no sleep. `durableSleep` tests already test the sleep.

- Fix `WorkflowWithStepAfterTimeout`: it was never identifying the workflow was cancelled because it used DBOS operations with a cancelled context itself. Fix the expectation on the error returned by the workflow

- Remove some `time.Sleep` calls for events

- `StreamWorkflowRecovery`: tolerate double step writes when `WriteStream` is called within a step.

- Add a missing db retry during recovery

### DIFF
--- a/chaos_tests/chaos_test.go
+++ b/chaos_tests/chaos_test.go
@@ -179,7 +179,10 @@ func PostgresChaosMonkey(t *testing.T, ctx context.Context, wg *sync.WaitGroup) 
 			// Check for context cancellation first
 			select {
 			case <-ctx.Done():
-				startPostgres(t, cliPath)
+				require.Eventually(t, func() bool {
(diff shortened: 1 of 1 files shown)


dbos/workflows_test.go
```diff
@@ -2747,113 +2766,6 @@ func durableGetEventSleepWorkflow(ctx DBOSContext, targetWorkflowID string) (str
 	return val1 + val2, nil
 }
 
-// Test workflows and steps for parameter mismatch validation
```
maxdml, line 2750: just moved under the get/set event tests

PR #245, opened 2026-01-31
Refactor old special steps

maxdml:
Move `setEvent` and `send` to be run throughout `runAsTxn`.

Ideally we'd run `getEvent`, `recv`, and `sleep` using `runAsTxn`, but this is currently very challenging because:
- `getEvent`/`recv` update the stepID on their own, to generate a step ID for `sleep`. This complicates the whole logic of checkpointing
- `sleep` has its own logic to check whether it executed or not, to become durable

Also:
- Allow `setEvent` to be sent within a step
- Allow retrying transactions during `runAsTxn` for CRDB
- Fix > 1 msg consumption in Recv
- Allow setting isolation level for `runAsTxn` (default read committed)

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1119,20 +1119,12 @@ func (s *sysDB) resumeWorkflow(ctx context.Context, input resumeWorkflowDBInput)
 						  WHERE workflow_uuid = $5`, pgx.Identifier{s.schema}.Sanitize())
 
 	if input.tx != nil {
-		_, err = input.tx.Exec(ctx, "SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")
-		if err != nil {
-			return fmt.Errorf("failed to set transaction isolation level: %w", err)
-		}
 		_, err = input.tx.Exec(ctx, updateStatusQuery,
 			WorkflowStatusEnqueued,
 			_DBOS_INTERNAL_QUEUE_NAME,
 			0,
 			time.Now().UnixMilli(),
 			input.workflowID)
-		_, err = input.tx.Exec(ctx, "SET TRANSACTION ISOLATION LEVEL READ COMMITTED")
-		if err != nil {
-			return fmt.Errorf("failed to set transaction isolation level: %w", err)
-		}
 	} else {
 		_, err = s.pool.Exec(ctx, updateStatusQuery,
 			WorkflowStatusEnqueued,
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1190,12 +1190,13 @@ type txn[R any] func(ctx context.Context, tx pgx.Tx) (R, error)
 
 // stepOptions holds the configuration for step execution using functional options pattern.
 type stepOptions struct {
-	maxRetries         int           // Maximum number of retry attempts (0 = no retries)
-	backoffFactor      float64       // Exponential backoff multiplier between retries (default: 2.0)
-	baseInterval       time.Duration // Initial delay between retries (default: 100ms)
-	maxInterval        time.Duration // Maximum delay between retries (default: 5s)
-	stepName           string        // Custom name for the step (defaults to function name)
-	preGeneratedStepID *int          // Pre generated stepID
+	maxRetries         int             // Maximum number of retry attempts (0 = no retries)
+	backoffFactor      float64         // Exponential backoff multiplier between retries (default: 2.0)
+	baseInterval       time.Duration   // Initial delay between retries (default: 100ms)
+	maxInterval        time.Duration   // Maximum delay between retries (default: 5s)
+	stepName           string          // Custom name for the step (defaults to function name)
+	preGeneratedStepID *int            // Pre generated stepID
+	txIsoLevel         *pgx.TxIsoLevel // Transaction isolation level for runAsTxn (nil = ReadCommitted)
 }
 
 // setDefaults applies default values to stepOptions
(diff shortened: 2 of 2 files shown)


dbos/system_database.go
```diff
@@ -1680,10 +1680,6 @@ func (s *sysDB) sleep(ctx context.Context, input sleepInput) (time.Duration, err
 		return 0, newStepExecutionError("", functionName, fmt.Errorf("workflow state not found in context: are you running this step within a workflow?"))
 	}
 
-	if wfState.isWithinStep {
-		return 0, newStepExecutionError(wfState.workflowID, functionName, fmt.Errorf("cannot call Sleep within a step"))
-	}
-
```
maxdml, line 1686: Lifted to pre-sysdb invocation

dbos/system_database.go
```diff
@@ -2147,10 +2083,6 @@ func (s *sysDB) recv(ctx context.Context, input recvInput) (*string, error) {
 		return nil, newStepExecutionError("", functionName, fmt.Errorf("workflow state not found in context: are you running this step within a workflow?"))
 	}
 
-	if wfState.isWithinStep {
-		return nil, newStepExecutionError(wfState.workflowID, functionName, fmt.Errorf("cannot call Recv within a step"))
-	}
```
maxdml, line 2152: lifted to pre-sysdb invocation

dbos/workflow.go
```diff
@@ -1022,7 +1022,7 @@ func (c *dbosContext) RunWorkflow(_ DBOSContext, fn WorkflowFunc, input any, opt
 		insertStatusResult, err = c.systemDB.insertWorkflowStatus(uncancellableCtx, insertInput)
 		if err != nil {
 			c.logger.Error("failed to insert workflow status", "error", err, "workflow_id", workflowID)
-			return err
+			return newWorkflowExecutionError(workflowID, fmt.Errorf("failed to insert workflow status: %w", err))
```
maxdml, line 1025: This actually is a small breaking change -- wrap the sysdb error into a yet-more-explanatory workflow error. (Hence the changes to use error.Is to look for a wrapped error in the entire error tree.)

dbos/workflow.go
```diff
@@ -1035,9 +1035,7 @@ func (c *dbosContext) RunWorkflow(_ DBOSContext, fn WorkflowFunc, input any, opt
 				stepID:           parentWorkflowState.stepID,
 				tx:               tx,
 			}
-			err = retry(uncancellableCtx, func() error {
-				return c.systemDB.recordChildWorkflow(uncancellableCtx, childInput)
-			}, withRetrierLogger(c.logger))
+			err = c.systemDB.recordChildWorkflow(uncancellableCtx, childInput)
```
maxdml: Undeeded retry within a larger retry + was buggy because retry cancel was uncancellable.

dbos/system_database.go
```diff
@@ -2042,65 +2038,30 @@ type WorkflowSendInput struct {
...
-	// Get workflow state from context (optional for Send as we can send from outside a workflow)
-	wfState, ok := ctx.Value(workflowStateKey).(*workflowState)
-	var stepID int
-	var isInWorkflow bool
-
-	if ok && wfState != nil {
-		isInWorkflow = true
-		if wfState.isWithinStep {
-			return newStepExecutionError(wfState.workflowID, functionName, fmt.Errorf("cannot call Send within a step"))
-		}
```
maxdml, line 2062: Lifted to pre-sysdb invocation

dbos/debouncer.go
```diff
@@ -196,6 +197,19 @@ func (d *Debouncer[P, R]) Debounce(ctx DBOSContext, key string, delay time.Durat
...
+				var pgErr *pgconn.PgError
+				if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.SerializationFailure {
+					debouncerWorkflowStatus, reErr := ListWorkflows(ctx, WithFilterDeduplicationID(key))
+					if reErr != nil {
+						return nil, reErr
+					}
+					if len(debouncerWorkflowStatus) == 0 || debouncerWorkflowStatus[0].Status != WorkflowStatusPending {
+						continue
+					}
+				}
```
maxdml: @kraftp not handled in other languages either. When workflow status gets updated, Send() can fail with a 40001 and we exit Debounce -- we should just continue the loop to start a new debouncer workflow and not lose the call.


Image

dbos/system_database.go
```diff
@@ -2261,19 +2193,18 @@ loop:
...
-            FROM %s.notifications
-            WHERE destination_uuid = $1 AND topic = $2
-            ORDER BY created_at_epoch_ms ASC
-            LIMIT 1
-        )
-        DELETE FROM %s.notifications
-        WHERE destination_uuid = (SELECT destination_uuid FROM oldest_entry)
-          AND topic = (SELECT topic FROM oldest_entry)
-          AND created_at_epoch_ms = (SELECT created_at_epoch_ms FROM oldest_entry)
-        RETURNING message`, pgx.Identifier{s.schema}.Sanitize(), pgx.Identifier{s.schema}.Sanitize())
```
maxdml, line 2276: This query can delete more than 1 message if `created_at_epoch_ms` is within the same millisecond. This has been surfaced by 1) not running send() inside a transaction outside of a workflow and 2) a recent change to the Golang migration where we now have:

`created_at_epoch_ms BIGINT NOT NULL DEFAULT (EXTRACT(epoch FROM now())::numeric * 1000)::bigint,`

instead of

`created_at_epoch_ms BIGINT NOT NULL DEFAULT (EXTRACT(epoch FROM now()) * 1000.0)::bigint,
`

The first line converts the return value of `now()` (double precision, float8) to a numeric, which result in the `*1000` multiplication being exact numeric and quite stable.

The second line (what we have in Python) does the multiplication on double precision, then converts to bigint, which can do a truncation. Because multiplication on floating points can (often) have errors, this meant more volatility in the truncation, which, I think could have contributed to obfuscate this bug.

dbos/debouncer.go
kraftp: The core workflow operations run READ COMMITTED and can't throw 40001 exceptions. If they could, that would be annoying to users as they'd have to handle random serialization failures. Is Go's implementation different? If so, we should rethink it.

maxdml: I see. Right now I set `REPEATABLE READ` to all the queries ran through `runAsTxn`. I can move that to specific sys DB operations. We should sync up on which ones.

PR #250, opened 2026-02-05
Prevent concurrent wf execution

### LINKED ISSUE #211: Prevent concurrent execution of workflows / steps
Today it is possible for one (or more) processes to execute the same workflow concurrently. Albeit DBOS has mechanisms in place to enforce consistent behavior across concurrent executions, it still allows for undesirable effects (e.g., double the load on some API endpoints)

TS version: https://github.com/dbos-inc/dbos-transact-ts/pull/1120
Python version: https://github.com/dbos-inc/dbos-transact-py/pull/545

maxdml:
- Use an in-memory map to track if a workflow is being executed by a `dbosCtxt` instance
- Add `owner_xid` in the database as a distributed lock
- Only increment attempts counter when executing the first time, recovering or dequeuing
- Update tests to use recovery path for testing idempotency
- Remove some obsolete tests (e.g. Worker concurrency X idempotency)

Fixes #211

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1871,7 +1871,9 @@ func (s *sysDB) notificationListenerLoop(ctx context.Context) {
 
 	s.logger.Debug("DBOS: Starting notification listener loop")
 
-	poolConn, err := acquire(ctx)
+	poolConn, err := retryWithResult(ctx, func() (*pgxpool.Conn, error) {
+		return acquire(ctx)
+	}, withRetrierLogger(s.logger))
 	if err != nil {
 		s.logger.Error("Failed to acquire listener connection", "error", err)
 		return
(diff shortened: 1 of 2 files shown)


dbos/workflows_test.go
```diff
@@ -1541,39 +1471,77 @@ func TestWorkflowIdempotency(t *testing.T) {
 	})
 }
 
+func TestNoConcurrentWorkflowSameID(t *testing.T) {
```
maxdml, line 1474: new test

maxdml:
> Awesome work Max! It looks correct unless I missed something and the new tests are good. I'm fixing this in Python right now, but did you already fix/test in Go how getResult interacts with MAX_RECOVERY_ATTEMPTS_EXCEEDED?

yep!

PR #260, opened 2026-03-02
change log: workflow deletion with its children

jerinthomas1404:
- add delete workflow with its children to DBOS context with default option as false
- add getWorkflow children to fetch children associated to workflow id
- add workflow deletion tests: simple, with child, pending state, completed state, non-existent
- add delete workflow to CLI tests

### DIFF
--- a/dbos/client_test.go
+++ b/dbos/client_test.go
@@ -841,7 +841,7 @@ func TestForkWorkflow(t *testing.T) {
 			assert.Equal(t, expectedEventTuples, actualEventTuples, "forked workflow at step %d: events history mismatch", startStep)
 
 			// 3) Verify counters are at expected totals based on the step where we're forking
-			t.Logf("Step %d: actual counters - step1:%d, step2:%d, child1:%d, child2:%d", startStep, stepCount1, stepCount2, child1Count, child2Count)
+			t.Logf("Start step %d: actual counters - step1:%d, step2:%d, child1:%d, child2:%d", startStep, stepCount1, stepCount2, child1Count, child2Count)
 
 			expectedStep1Count := 1 + min(startStep+1, 5)
 			assert.Equal(t, expectedStep1Count, stepCount1, "forked workflow at step %d: step1 counter should be %d", startStep, expectedStep1Count)
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -49,7 +49,6 @@ type systemDatabase interface {
 	getWorkflowChildren(ctx context.Context, input getWorkflowChildrenDBInput) ([]WorkflowStatus, error)
 	recordChildWorkflow(ctx context.Context, input recordChildWorkflowDBInput) error
 	checkChildWorkflow(ctx context.Context, workflowUUID string, functionID int) (*string, error)
-	recordChildGetResult(ctx context.Context, input recordChildGetResultDBInput) error
 
 	// Steps
 	recordOperationResult(ctx context.Context, input recordOperationResultDBInput) error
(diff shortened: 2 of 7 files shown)


dbos/system_database.go
```diff
@@ -1433,6 +1435,59 @@ func (s *sysDB) recordOperationResult(ctx context.Context, input recordOperation
...
+		} else {
+			_, err = s.pool.Exec(ctx, deleteQuery, workflowID)
+		}
+		if err != nil {
+			return fmt.Errorf("failed to delete workflow: %w", err)
+		}
+	}
+
+	return nil
+}
```
maxdml: This should be a single query. No need to do a list workflows to check existence.

dbos/system_database.go
```diff
@@ -1433,6 +1435,59 @@ func (s *sysDB) recordOperationResult(ctx context.Context, input recordOperation
...
+	tx         pgx.Tx
+}
+
+// getWorkflowChildren retrieves all direct child workflows of the given parent workflow.
+func (s *sysDB) getWorkflowChildren(ctx context.Context, input getWorkflowChildrenDBInput) ([]WorkflowStatus, error) {
+	return s.listWorkflows(ctx, listWorkflowsDBInput{
+		parentWorkflowID: []string{input.workflowID},
+		tx:               input.tx,
+	})
+}
```
maxdml: The recursive logic should be better placed here. Also the entire search should be done in the same transaction. (No point in having this function that's just a pass through for `listWorkflows`.)

dbos/workflows_test.go
```diff
@@ -5368,3 +5368,117 @@ func collectStreamValues[R any](ch <-chan StreamValue[R]) ([]R, bool, error) {
 
 	return values, closed, err
 }
+
+func TestWorkflowDelete(t *testing.T) {
+	dbosCtx := setupDBOS(t, setupDBOSOptions{dropDB: true, checkLeaks: true})
```
maxdml: Let's place this test in the `TestChildWorkflow` suite, which already includes declaration for parents & grand parents. Let's also test with a more complex topology (3 layers graph with at least 2 nodes in each layer)

dbos/system_database.go
jerinthomas1404: - Okay, I will make it to a single query.
- So after getting the child workflow IDs, I should delete all workflows that are NOT in an active state, right?

maxdml: You can just delete all of them, regardless of their state. The point of this API is to be a definitive destroy button on specific workflows.

jerinthomas1404: Got it. I will raise the PR with these changes by EOD.

dbos/system_database.go
jerinthomas1404: I have moved the recursive logic.

dbos/workflows_test.go
jerinthomas1404: - Done

dbos/workflows_test.go
```diff
@@ -1325,6 +1325,53 @@ func TestChildWorkflow(t *testing.T) {
...
+				return "", err
+			}
+			if _, err := h.GetResult(); err != nil {
+				return "", err
+			}
+		}
+		return "root:" + input, nil
+	}
+	RegisterWorkflow(dbosCtx, deleteRootWf)
+	t.Cleanup(func() { deleteBlockEvent.Set() })
```
maxdml, line 1389: Let's add some events, notifications and streams to this workflow. The test must verify that these have been deleted alongside the workflow (this should happen because of the `ON DELETE CASCADE` clause on the `workflow_status` table.)

dbos/client_test.go
maxdml, line 1: We can make this simpler -- no need to repeat all the tests. Just do a simple one to validate the client methods.

dbos/workflows_test.go
jerinthomas1404: - Okay, I'll pick it up.

jerinthomas1404:
@maxdml , I have made the changes for test.

dbos/workflows_test.go
```diff
@@ -1454,6 +1517,149 @@ func TestChildWorkflow(t *testing.T) {
...
+		require.ErrorAs(t, err, &dbosErr)
+		require.Equal(t, NonExistentWorkflowError, dbosErr.Code)
+	})
+
+	t.Run("DeleteNonExistentWorkflowIsNoOp", func(t *testing.T) {
+		err := DeleteWorkflow(dbosCtx, "non-existent-delete-wf-id")
+		require.NoError(t, err, "expected no error when deleting non-existent workflow")
+	})
+
+	t.Run("DeleteCascadesRelatedData", func(t *testing.T) {
```
maxdml, line 1562: This should also test that the step outcomes (GetWorkflowSteps) have been cascade deleted.

dbos/workflows_test.go
```diff
@@ -1325,6 +1325,69 @@ func TestChildWorkflow(t *testing.T) {
...
+			if _, err := h.GetResult(); err != nil {
+				return "", err
+			}
+		}
+		return "root:" + input, nil
+	}
+	RegisterWorkflow(dbosCtx, deleteRootWf)
+
+	// Workflow for cascade data deletion test
+	deleteCascadeWf := func(ctx DBOSContext, _ string) (string, error) {
```
maxdml, line 1375: This workflow should call Recv() to get the notification you send in the test (recv is recorded as a special step in the steps table)

dbos/workflow.go
```diff
@@ -2826,6 +2826,81 @@ func CancelWorkflow(ctx DBOSContext, workflowID string) error {
...
+// events, event history, and streams associated with the workflow.
+//
+// Options:
+//   - WithDeleteChildren: Also delete all child workflows recursively
+//
+// Parameters:
+//   - ctx: DBOS context for the operation
+//   - workflowID: The unique identifier of the workflow to delete
+//
+// Returns an error if the workflow does not exist, is still active, or if the deletion fails.
```
maxdml, line 2888: Correct this. We don't return an error if the workflow is still active.

PR #270, opened 2026-03-15
Complete mock coverage for all DBOSContext interface methods   

azghr:
1. Added WithValue and WithCancelCause as interface methods (previously standalone functions)
2. Implemented methods on dbosContext to enable mocking                                                     
3. Added mock tests for remaining untested methods                                                          

  All DBOSContext methods now have mock test coverage.

### DIFF
--- a/integration/mocks_test.go
+++ b/integration/mocks_test.go
@@ -173,6 +173,88 @@ func aRealProgramFunction(dbosCtx dbos.DBOSContext) error {
 	return nil
 }
 
+// clientMethodsFunction exercises remaining DBOSContext methods
+func clientMethodsFunction(ctx dbos.DBOSContext) error {
+	// WriteStream
+	err := dbos.WriteStream(ctx, "stream-key", "stream-value")
+	if err != nil {
+		return fmt.Errorf("WriteStream failed: %w", err)
+	}
+
+	// CloseStream
+	err = dbos.CloseStream(ctx, "stream-key")
+	if err != nil {
+		return fmt.Errorf("CloseStream failed: %w", err)
+	}
+
+	// ReadStream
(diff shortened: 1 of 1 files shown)


integration/mocks_test.go
```diff
@@ -188,8 +188,89 @@ func TestMocks(t *testing.T) {
...
+	mockNoCancelCtx := mocks.NewMockDBOSContext(t)
+	mockCtx2.On("WithoutCancel", mockCtx2).Return(mockNoCancelCtx).Maybe()
+
+	// WithTimeout
+	mockTimeoutCtx := mocks.NewMockDBOSContext(t)
+	var cancel context.CancelFunc = func() {}
+	mockCtx2.On("WithTimeout", mockCtx2, mock.Anything).Return(mockTimeoutCtx, cancel).Maybe()
+
+	// ListenQueues
+	mockCtx2.On("ListenQueues", mockCtx2, mock.Anything).Maybe()
```
maxdml: This needs a real function using the client, that we can mock. The checks shouldn't be `Maybe` but actually check the method was called as expected, with specific input values.

maxdml (commented):
Thanks @azghr. We also need to update the `workflow` function so that it also calls missing `DBOSContext` interface methods (e.g., `Select`)

integration/mocks_test.go
azghr: I address the changes... please review @maxdml

maxdml: @azghr Thanks! Can you make sure the tests pass please?

azghr: it's done... now all the GitHub Actions are passing. @maxdml

PR #295, opened 2026-04-14
delayed execution

maxdml:


devhawk:
I think `setWorkflowDelay` is missing from `Client` and `DBOSContext`

dbos/queue.go
```diff
@@ -254,6 +254,11 @@ func (qr *queueRunner) runQueue(ctx *dbosContext, queue WorkflowQueue) {
 		hasBackoffError := false
 		skipDequeue := false
 
+		// Transition any DELAYED workflows whose delay has expired to ENQUEUED.
+		if err := ctx.systemDB.transitionDelayedWorkflows(ctx); err != nil {
```
devhawk, line 258: Are we supposed to be calling `transitionDelayedWorkflows` once per polling interval *per queue*? That seems incorrect. Python calls it once per polling interval in `queue_thread`, but individual `queue_worker_thread` are spun up from `queue_thread`.

maxdml: Right now the architecture only has "living" loops for each queue, unlike Python. When we add the queue metadata in the database, I'll make this optimization (because queues will be dynamic too.)

maxdml:
> I think `setWorkflowDelay` is missing from `Client` and `DBOSContext`

Fixed

kraftp (approved):

PR #298, opened 2026-04-16
Implement database-backed workflow scheduling (Schedules 2.0)

azghr:
Add Workflow schedule management API (create, pause, resume, delete, list, backfill, trigger).

Call `CreateSchedule(_ DBOSContext, fn ScheduledWorkflowFunc, input CreateScheduleRequest, opts ...CreateScheduleOption) error` to create a schedule from a workflow function registerd with DBOSContext.

The workflow signature is `type ScheduledWorkflowFunc func(ctx DBOSContext, input ScheduledWorkflowInput) (any, error)` and must accept:

```go
type ScheduledWorkflowInput struct {
	ScheduledTime time.Time `json:"scheduled_time"`
	Context       any       `json:"context,omitempty"`
}
```

Where `context` can be any serializable data the workflow can access at runtime. Typecasting is left to the user.

This still rely on [robfig/cron](https://github.com/robfig/cron). Not too hard to replace if we need to, down the line.

_(fixes: also store serialization method in import/export and fix a race in Shutdown())_


maxdml:
@azghr thanks for the PR. I'll take it from here.

azghr:
> @azghr thanks for the PR. I'll take it from here.

Thank you @maxdml ... I think this is the right direction

dbos/client.go
```diff
@@ -563,6 +576,274 @@ func ClientReadStreamAsync[R any](c Client, workflowID string, key string) (<-ch
...
+	contextJSON, err := json.Marshal(input.Context)
+	if err != nil {
+		return fmt.Errorf("failed to serialize context: %w", err)
+	}
+
+	return dbosCtx.systemDB.createSchedule(dbosCtx, createScheduleDBInput{
+		ScheduleID:        scheduleID,
+		ScheduleName:      input.ScheduleName,
+		WorkflowName:      input.WorkflowName,
+		WorkflowClassName: input.WorkflowClassName,
```
maxdml, line 627: Class name is not used in Go but useful for a Go Client ot apply schedules for another language's worker.

dbos/workflow.go
```diff
@@ -1474,13 +1476,6 @@ func WithNextStepID(stepID int) StepOption {
 	}
 }
 
-// withTxIsolationLevel sets the transaction isolation level for runAsTxn (package-private).
-func withTxIsolationLevel(level pgx.TxIsoLevel) StepOption {
-	return func(opts *stepOptions) {
-		opts.txIsoLevel = &level
-	}
-}
```
maxdml, line 1482: unused

kraftp (approved):
Nice work! As far as I can tell, features are at parity and we have tests for the important edge cases (particularly around time zones and backfill).

PR #313, opened 2026-05-14
application versions management

maxdml:
Also add a few missing DB retries


kraftp (commented):
What about the client?

kraftp (approved):

PR #315, opened 2026-05-14
Partial indexes

maxdml:
Also add intermediate, missing migrations.

### DIFF
--- a/dbos/application_versions_test.go
+++ b/dbos/application_versions_test.go
@@ -1,0 +1,90 @@
+package dbos
+
+import (
+	"errors"
+	"testing"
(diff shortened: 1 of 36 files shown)


dbos/application_versions_test.go
```diff
@@ -0,0 +1,90 @@
+package dbos
```
kraftp, line 1: Git history looks screwed up here

maxdml: Should be fixed after the merge!

kraftp (approved):
Looks good! Let's just make sure the migrations match exactly

PR #317, opened 2026-05-16
Sqlite

### LINKED ISSUE #304: Add SQLite support
Had another request for this over the weekend.

maxdml:
Add SQLite support for Transact Go.

This entails hidding specific backends specificities behind a driver-agnostic adapter interface (`pgxPoolAdapter`, `newSQLPool`) and a dialect interface. The adapters are stateful, host the underlying pool. The dialects are stateless and invoked to resolve query formatting, etc.

```json
Usage: Give a sqlite databaseURL or pass a *sql.DB pointer to dbos.Config

e.g., sqlite:/tmp/dbos.db
```

We apply the following defaults to SQLite:
- `_txlock=immediate`
- `busy_timeout(5000)`
- `foreign_keys(ON)`
- `synchronous(NORMAL)`
- Max open connections: 8
- Max idle connections: 4
- Connection max lifetime: 1h

**Details**

We use [modernc.org/sqlite](https://modernc.org/sqlite), a lightweight alternative to [mattn/go-sqlite3](https://github.com/mattn/go-sqlite3).

- The dialect interface handle things like schema formatting or Query formatting (e.g., formatting array parameters)
- SQLite bootstraping part is not shared with Postgres/CockRoach. Too specific.
- Migrations: the SQLite migrations don't use `now_epoch_ms()` and `gen_random_uuid()`, so we don't have to attached them as UDF.
- `cancelWorkflows` and `resumeWorkflows` uses Data-modifying CTEs, which sqlite doesn't support, so different code path to do write + read in a txn.

Closes #304


kraftp (commented):
I've been thinking about this for a little while. The approach with the dialect adapter make sense on paper, and all the tests pass so it's probably mostly correct. I'm not completely comfortable with a thousand lines of hard-to-review AI-generated code on the critical path of everything, but I don't see a better alternative for supporting SQLite.

Another question to consider is what the default database should be. In Python, we made SQLite the default to make it easier to get started with DBOS. As far as I can tell, this PR doesn't change the Go default--should it?

maxdml:
Yeah, given that we use a driver and not an ORM (which is a good thing imo), we must add a layer to talk the right lingua to each driver. I did review everything -- not that this is bullet-proof, but I'm confident it works. The parts I am the most skeptical about, because I don't have experience operating sqlite, are the connection parameters we set. These are based on a little bit of research.

With respect to making sqlite the default, I'd keep Postgres the default and switch to sqlite in v1.

kraftp (approved):

PR #318, opened 2026-05-18
Fix goroutine leak when ReadStreamAsync consumer stops reading

merlinfuchs:
When `ReadStreamAsync` is called but the consumer stops reading from the channel before the stream is closed, the goroutine currently blocks forever on `ch <- v` without respecting context cancellation. 

You can verify that this bug exists by running the `GoroutineLeakOnContextCancel` test on `main`.

This PR introduces a small `send` wrapper that selects between `ch <- v` and `<-c.Done()` and immediately exits the loop when the context is done.


maxdml (approved):
Great fix, thanks @merlinfuchs

PR #319, opened 2026-05-18
Add DBOSContext.WithCancel

merlinfuchs:
Seems like this was left out unintentionally as `WithTimeout` and `WithCancelCause` already exist.


maxdml (approved):
Thanks @merlinfuchs

PR #320, opened 2026-05-18
Add DBOSContext.ReadStreamSnapshot method

merlinfuchs:
Right now there is no way to read workflow events without either blocking forever or a creating a Goroutine the writes to a channel until the stream is closed. 

For some scenarios it would be useful to have a way to retrieve only the events that have already been written to a stream without blocking. `ReadStreamSnapshot` allows this and also adds the ability to only read events after a certain offset.

Not sure if this is the right approach. An alternative would be to repurpose `ReadStream` for this and accept options like `dbos.WithWait(false)` and `dbos.WithOffset` 🤔 

Based on #318

### DIFF
--- a/dbos/client.go
+++ b/dbos/client.go
@@ -546,42 +546,56 @@ func ClientReadStreamAsync[R any](c Client, workflowID string, key string) (<-ch
 	}
 
 	typedCh := make(chan StreamValue[R], 1)
+	dbosCtx := c.(*client).dbosCtx
 
 	go func() {
 		defer close(typedCh)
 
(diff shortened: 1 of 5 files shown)


maxdml:
@merlinfuchs Thanks for the PR. I'd prefer repurposing `ReadStream` to block by default and be configured to return after the first read (e.g., `WithReadStreamSnapshot`) and take an optional offset.

merlinfuchs:
@maxdml Agreed, I will make a new PR for this :)

merlinfuchs:
closing in favor of #322

PR #321, opened 2026-05-19
Add missing enqueue options

merlinfuchs:
Adds missing options for `AuthenticatedUser`, `AssumedRole` and `AuthenticatedRoles` to `dbos.Enqueue`. While these fields aren't used anywhere in DBOS I was planning to utilize them myself to attach some metadata to a workflow.

The options are already available on `RunWorkflow` so this PR just closes the gap.

### DIFF
--- a/dbos/client.go
+++ b/dbos/client.go
@@ -176,6 +176,27 @@ func WithEnqueueDelay(delay time.Duration) EnqueueOption {
 	}
 }
 
+// WithEnqueueAuthenticatedUser sets the authenticated user for the enqueued workflow.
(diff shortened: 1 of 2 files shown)


maxdml:
Thanks @merlinfuchs , let's just add a small check on the unit tests to verify they are passed through correctly

merlinfuchs:
@maxdml Added a test :)

maxdml (approved):

maxdml:
@merlinfuchs the crdb test failure is unrelated, should be fixed in #317

PR #323, opened 2026-05-20
Remove log noise for unknown schedule workflows

merlinfuchs:
I was planning to run multiple workers with different workflows all connected to the same DBOS database + schema. This works great but as soon as I create schedules, each worker starts logging `failed to get workflow for schedule` for any workflow they don't know about.

Maybe the proper fix would be for each worker to only look at the schedules for queues and workflows they have registered? For now I just lowered the log level to debug because I wasn't sure about the intended design.

### DIFF
--- a/dbos/scheduler.go
+++ b/dbos/scheduler.go
@@ -214,7 +214,7 @@ func (c *dbosContext) buildDBScheduleFunc(schedule WorkflowSchedule) (ScheduledW
 func (c *dbosContext) addDBScheduleToScheduler(schedule WorkflowSchedule) {
 	fn, err := c.buildDBScheduleFunc(schedule)
 	if err != nil {
-		c.logger.Error("failed to get workflow for schedule", "schedule", schedule.ScheduleName, "error", err)
+		c.logger.Debug("failed to get workflow for schedule", "schedule", schedule.ScheduleName, "error", err)
 		return
 	}
 


maxdml:
@merlinfuchs are these effectively different applications wired to the same DBOS system database? We'd recommend using different system database for this.

The reason why this log is an error is to surface a very unexpected behavior, following the model above.

merlinfuchs:
Sorry for the late reply! It's one application split across multiple (micro-)services. Would you still recommend having separate DBOS databases / schemas for each service? That makes it less easy for service A to enqueue a job for service B without having a DBOS client for each service ...

maxdml:
> Sorry for the late reply! It's one application split across multiple (micro-)services. Would you still recommend having separate DBOS databases / schemas for each service? That makes it less easy for service A to enqueue a job for service B without having a DBOS client for each service ...

DBOS works best with 1 DBOS app per application. The service architecture is, typically, 1 service <-> 1 API <-> 1 database. May I ask: what are the services in your case?

And, yes, the solution for services are clients.

chenjpu:
We are facing the same issue with multiple services sharing one schema.

maxdml:
@merlinfuchs @chenjpu 

One thing you can do today, is setup "explicit queue listening" before launching the DBOS context.

What this does is have the local DBOS queue runner dequeue from a select list of queues only.

You'd have to wire that to something the environment expose, e.g., an env variable that you can use to dynamically decide what queues to listen to.

https://docs.dbos.dev/golang/reference/queues#listenqueues

AminMortezaie:
Thanks for the detailed feedback on the intended architectural boundaries of DBOS. I completely understand the design philosophy here: DBOS excels at providing lightweight, embedded orchestration without the operational tax of a heavy external cluster like Temporal, making it incredibly powerful for monoliths and single-process applications (like AI agents).
However, looking at how startups scale, there is a major architectural friction point when they begin breaking their applications down into microservices. While '1 service <-> 1 database' is the ideal microservice standard, forcing services to completely decouple means they lose the ability to easily orchestrate cross-service flows without managing complex multi-database clients or building manual HTTP/REST wrapper logic. Conversely, sharing the same DBOS database schema (as attempted here) breaks encapsulation and introduces log noise due to the global cron table polling mechanics.
To prevent teams from outgrowing DBOS and migrating to Temporal the moment they scale past a monolith, I believe the framework needs a structured path to handle microservice environments.
Instead of a centralized orchestrator cluster, we could introduce features that support a Federated/Decentralized Mesh Architecture:

   1. Intelligent Schedule/Queue Filtering: At a minimum, updating the internal background runners so they only fetch or lock table rows for workflows that were explicitly registered via dbos.RegisterWorkflow() at application startup. This would immediately eliminate log noise and runtime errors in multi-worker or multi-service setups sharing a backend store.
   2. Abstracted Communication Layers: Long-term, if the core WorkflowContext could support a light transport abstraction (like gRPC or NATS) for cross-boundary hand-offs (e.g., an InvokeRemoteStep), DBOS could coordinate states between independent microservices without requiring them to share a physical database.

I would love to help contribute toward making DBOS more resilient for scaling architectures. Would the maintainers be open to a PR that introduces smart workflow/schedule filtering based on locally registered types to resolve this log noise issue?"

maxdml:
> To prevent teams from outgrowing DBOS and migrating to Temporal the moment they scale past a monolith, I believe the framework needs a structured path to handle microservice environments.

This is totally unrelated, I'm closing this PR at this point

chenjpu:
@maxdml, would it be acceptable to submit a PR to support service isolation by adding WorkflowNames and ScheduleNamePrefixes configuration fields to the scheduler?

**Implementation:**
This would involve passing these filter parameters into the listSchedulesDBInput during the reconciliation loop:
```go
// Current code (scheduler.go#L282)
schedules, err := c.systemDB.listSchedules(c, listSchedulesDBInput{})

// Proposed change
schedules, err := c.systemDB.listSchedules(c, listSchedulesDBInput{
    WorkflowNames:        c.config.WorkflowNames,
    ScheduleNamePrefixes: c.config.ScheduleNamePrefixes,
})
````

**Benefits**
  1.Eliminates Log Noise: Services will no longer process or throw errors for unknown workflows.
  2.Microservices Support: Provides a lightweight, non-intrusive isolation mechanism for multi-service architectures sharing a single database.
  3.Backward Compatible: If these configuration fields are left empty, the behavior remains identical to the current implementation.

maxdml:
@chenjpu that's a good idea. We'll chat about it today.

maxdml:
@chenjpu what you can already do, today, is scheduling to a queue which is explicitly listened to by some executors.

We prefer having a queue-centric model for worker-allocation rather than mixing it between queues and schedules.

So you'd do:
* Use `WithScheduleQueueName` when calling `CreateSchedule` (I'll update the docs soon)
* Then [ListenQueues](https://docs.dbos.dev/golang/reference/queues#listenqueues)

chenjpu:
> @chenjpu what you can already do, today, is scheduling to a queue which is explicitly listened to by some executors.
> 
> We prefer having a queue-centric model for worker-allocation rather than mixing it between queues and schedules.
> 
> So you'd do:
> 
> * Use `WithScheduleQueueName` when calling `CreateSchedule` (I'll update the docs soon)
> * Then [ListenQueues](https://docs.dbos.dev/golang/reference/queues#listenqueues)

I followed the previous suggestions, but the error logs are still showing up. Did I miss a specific configuration parameter?

maxdml:
@chenjpu good catch. I'll fix it for the next release.

PR #324, opened 2026-05-29
WithReadStreamSnapshot

maxdml:


dbos/dbos.go
```diff
@@ -153,8 +153,8 @@ type DBOSContext interface {
 	GetEvent(_ DBOSContext, targetWorkflowID string, key string, timeout time.Duration) (any, error)            // Get a key-value event from a target workflow
 	WriteStream(_ DBOSContext, key string, value any, opts ...WriteStreamOption) error                          // Write a value to a durable stream
 	CloseStream(_ DBOSContext, key string) error                                                                // Close a durable stream
-	ReadStream(_ DBOSContext, workflowID string, key string) ([]any, bool, error)                               // Read values from a durable stream (blocks until workflow inactive or stream closed)
-	ReadStreamAsync(_ DBOSContext, workflowID string, key string) (<-chan StreamValue[any], error)              // Read values from a durable stream asynchronously
+	ReadStream(_ DBOSContext, workflowID string, key string, opts ...ReadStreamOption) ([]any, bool, error)                 // Read values from a durable stream (blocks until workflow inactive or stream closed; WithReadStreamSnapshot drains available values and returns)
```
kraftp: I don't fully understand this interface. Doesn't `ReadStream` always return a snapshot? What happens if you pass `snapshot` into `ReadStreamAsync`, what's the point of it being async then?

maxdml: - `ReadStream` is blocking
- `ReadStreamAsync` returns you a channel (which technically you could read only partially, but not very user friendly)
- `ReadStreamSnapshot` just returns you a set of values at the time of the reading (not a channel)

maxdml: Good point: `ReadStreamAsync` does not need to accept the snapshot option (that's actually confusing)

kraftp (approved):

PR #327, opened 2026-06-03
refactor: use crdb_version parameter to detect CockroachDB

### LINKED ISSUE #326: CockroachDB detection requires elevated privileges
## Problem

The current CockroachDB detection logic relies on executing:

```sql
SHOW CLUSTER SETTING version
```

to determine whether the connected database is a CockroachDB cluster.

```go
func isCockroachDB(ctx context.Context, conn *pgx.Conn) bool {
    var version string
    err := conn.QueryRow(ctx, "SHOW CLUSTER SETTING version").Scan(&version)
    return err == nil
}
```

However, `SHOW CLUSTER SETTING version` may require elevated privileges depending on the cluster configuration and user permissions. As a result, database type detection can fail when the application is running with a restricted service account, even though the target database is CockroachDB.

## Expected Behavior

Database type detection should not depend on cluster-level permissions and should work with standard application database accounts.

## Proposed Solution

Use the PostgreSQL protocol startup metadata exposed by CockroachDB instead of executing a SQL query.

CockroachDB provides the `crdb_version` parameter through the connection metadata, which can be accessed via:

```go
conn.PgConn().ParameterStatus("crdb_version")
```

This approach:

* Eliminates the need for an additional SQL query
* Avoids privilege-related failures
* Reduces startup overhead
* Uses a CockroachDB-specific identifier exposed by the wire protocol

## Acceptance Criteria

* CockroachDB detection no longer executes `SHOW CLUSTER SETTING version`
* Detection works with non-admin database accounts
* PostgreSQL connections continue to be correctly identified as non-CockroachDB
* Existing functionality remains unchanged

chenjpu:
fix https://github.com/dbos-inc/dbos-transact-golang/issues/326

### DIFF
--- a/chaos_tests/chaos_test.go
+++ b/chaos_tests/chaos_test.go
@@ -53,9 +53,7 @@ func (e *Event) Set() {
 }
 
 func isCockroachDB(ctx context.Context, conn *pgx.Conn) bool {
-	var version string
-	err := conn.QueryRow(ctx, "SHOW CLUSTER SETTING version").Scan(&version)
-	return err == nil
+	return conn.PgConn().ParameterStatus("crdb_version") != ""
 }
 
 // dropDatabaseIfExists drops a database in a way that works with both PostgreSQL and CockroachDB.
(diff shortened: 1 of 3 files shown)


maxdml (approved):

maxdml:
Thanks @chenjpu . Fixing the vulncheck in #328

Issue #330, opened 2026-06-04
High `xact_rollback` rate


denpeshkov:
Hi team!

We are using DBOS v0.12.0 with PostgreSQL 18.

Our otherwise-idle DBOS workers (3 workers plus one client used only for enqueuing) produce a steady stream of rolled-back transactions visible in `pg_stat_database` and our Grafana dashboard.
We are currently seeing a rollback-to-total-transactions ratio of up to 80%.

I traced this to the queue runner: every ~1s an empty dequeue opens a transaction and rolls it back instead of committing. The comment in the dequeue SQL explains the intent (avoiding WAL bloat / XID advancement). `checkOperationExecution` does the same thing for step lookups.

A few questions:

1. Is this the intended behavior, or did I misconfigure something on my side?
2. AFAIK PostgreSQL uses a virtual XID for read-only transactions — would it be possible to use `BEGIN READ ONLY` instead?

Thank you!

maxdml:
@denpeshkov I'll be implementing a few queue optimizations that'll fix this soon.

BTW the latest version is v0.16.0 -- the queue improvements will be part of v0.17.0

PR #337, opened 2026-06-04
track worker queue tasks being executed in memory

### LINKED ISSUE #299: Use in-memory map to enforce queue worker concurrency
Instead of requiring REPEATABLE READ on the dequeue transaction.

maxdml:
Closes #299 and #330

### DIFF
--- a/dbos/dialect.go
+++ b/dbos/dialect.go
@@ -70,6 +70,11 @@ type Dialect interface {
 	// RepeatableRead; SQLite returns Default (the IMMEDIATE BEGIN handles it).
 	SnapshotIsolation() IsoLevel
 
+	// QueueDequeueIsolation returns the IsoLevel for the queue dequeue
+	// transaction. snapshot=true requests snapshot semantics
+	// snapshot=false allows the lighter read-committed path.
+	QueueDequeueIsolation(snapshot bool) IsoLevel
+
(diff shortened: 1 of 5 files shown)


dbos/dialect.go
```diff
@@ -70,6 +70,11 @@ type Dialect interface {
 	// RepeatableRead; SQLite returns Default (the IMMEDIATE BEGIN handles it).
 	SnapshotIsolation() IsoLevel
 
+	// QueueDequeueIsolation returns the IsoLevel for the queue dequeue
+	// transaction. snapshot=true requests snapshot semantics
+	// snapshot=false allows the lighter read-committed path.
+	QueueDequeueIsolation(snapshot bool) IsoLevel
```
kraftp, line 76: What is this? What does it do in SQLite? Can't it just be called "Isolation Level" and take in an isolation level string (but even then, what does it do in sqlite)?

kraftp (approved):
Looks good other than the one comment!

dbos/dialect.go
maxdml: It's a helper function to decide whether to return read committed or repeatable read.
In SQLite it always returns `sql.LevelDefault`, because SQLite only supports `SERIALIZABLE`.

PR #338, opened 2026-06-04
More workflow filters

maxdml:
* completed at
* dequeued at
* is forked from

list workflow filters:
* complete after/before
* dequeued after/before
* is forked from
* has parents


kraftp (approved):
This looks good, but we'll need another PR for the aggregates endpoints

PR #341, opened 2026-06-06
Auth info from parent to child

### LINKED ISSUE #331: Flow auth info from parents to child
identity info are lost from parent to children

salmangada:
Changes

- When spawning a child workflow (`isChildWorkflow == true`), propagate the parent's auth fields into the child's `params` before `workflow_status` is inserted - only fills empty fields so explicit overrides are respected.
- Populate `wfState` with auth from `params` at workflow start so the running goroutine carries the identity for its own children.

Behaviour 
- Auth is set once at the initial `RunWorkflow` call site
- All child and grandchild workflows inherit it automatically with no changes to workflow business logic
- A child that passes its own `WithAuthenticatedUser(...) `opts uses those instead - parent auth does not override explicit values
- Recovery correctly restores the original identity so post-crash child spawns inherit the same auth as the first run.

Closes #331

### DIFF
--- a/dbos/recovery.go
+++ b/dbos/recovery.go
@@ -51,10 +51,15 @@ func recoverPendingWorkflows(ctx *dbosContext, executorIDs []string) ([]Workflow
 			continue
 		}
 
-		// Convert workflow parameters to options
+		// Convert workflow parameters to options.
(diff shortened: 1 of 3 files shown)


maxdml (approved):
Thanks @salmangada

PR #343, opened 2026-06-08
Add typed admin client and stabilize time filter test

ibrhmkoz:
## Summary

  - Add typed AdminClient.
  - Replace manual HTTP requests in the time-filtering test.
  - Fix flaky list endpoint time-filtering test.

  ## Flakiness

  Previous test used time.Now() between workflow creations.

  Workflow created_at uses millisecond precision. Fast executions could place the workflow and boundary in the same millisecond. Inclusive filters then returned unexpected workflows.

  New test:

  - Creates five workflows with distinct timestamps.
  - Uses persisted third workflow timestamp as boundary.
  - Verifies 3 + 2 = 5 workflows.

  ## Testing

```sh
DBOS_TEST_BACKEND=sqlite go test ./dbos \
    -run '^TestAdminServer/List_endpoints_time_filtering$' -count=30
```

```sh
DBOS_TEST_BACKEND=sqlite go test ./dbos \
    -run '^TestAdminServer$' -count=1
```

### DIFF
--- a/dbos/admin_client.go
+++ b/dbos/admin_client.go
@@ -1,0 +1,256 @@
+package dbos
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const defaultAdminClientTimeout = 5 * time.Second
+
+type AdminClient struct {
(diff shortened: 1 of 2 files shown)


maxdml:
Thanks @ibrhmkoz. We are in the process of deprecating the admin server -- likely coming in a 5 or 6 weeks, so that's not super relevant now.

PR #346, opened 2026-06-09
Feat : Custom retry conditions in steps

### LINKED ISSUE #302: Custom retry conditions
Allow users to pass a function, to a step, that will be called to decide whether the error is retryable.

salmangada:
Adds a `WithRetryPredicate(func(error) bool)` option to `RunAsStep` that lets callers decide per-error whether a retry should happen. If the predicate returns false, the step stops immediately without waiting through the remaining backoff - even if maxRetries has not been reached. When not set, all errors are retried as before with no behaviour change.

Closes #302

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -426,32 +426,6 @@ func retryPredicateWorkflow(ctx DBOSContext, _ string) (string, error) {
 	)
 }
 
-var retryPredicateAllowAllCount int
-
-// retryPredicateAllowAllWorkflow: predicate always returns true same as no predicate.
-func retryPredicateAllowAllWorkflow(ctx DBOSContext, _ string) (string, error) {
-	return RunAsStep(ctx, func(_ context.Context) (string, error) {
-		retryPredicateAllowAllCount++
-		return "", fmt.Errorf("always transient")
-	},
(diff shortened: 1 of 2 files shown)


dbos/workflows_test.go
```diff
@@ -405,6 +405,54 @@ func stepRetryAlwaysFailsStep(_ context.Context) (string, error) {
...
+// retryPredicateNilWorkflow: no predicate set - all errors retried up to maxRetries.
+func retryPredicateNilWorkflow(ctx DBOSContext, _ string) (string, error) {
+	return RunAsStep(ctx, func(_ context.Context) (string, error) {
+		retryPredicateNilCount++
+		return "", fmt.Errorf("always fails")
+	},
+		WithStepMaxRetries(2),
+		WithBaseInterval(1*time.Millisecond),
+	)
+}
```
maxdml: Let's simplify this a bit. We just need the `retryPredicateWorkflow` case (`retryPredicateNilWorkflow` is the default and exercised in other tests, `retryPredicateAllowAllWorkflow` is overkill)

salmangada: Noted @maxdml , this makes sense to keep only necessary cases here.

maxdml (approved):

Issue #345, opened 2026-06-09
RunWorkflow and registry can execute different workflow functions


ibrhmkoz:
RunWorkflow(ctx, fn, input) resolves fn in the workflow registry, but direct execution calls the provided fn. Queued execution and recovery call the function captured during RegisterWorkflow.

Current behavior:

- Direct execution: provided fn
- Queue/recovery: registered fn

Different function values can resolve to the same FQN while having different captured state or behavior, such as closures from the same expression or methods bound to different receivers.

Calling RunWorkflow with such a function may directly execute a different workflow body than the registered body. The implementations can have different steps, step ordering, branching, or results. Recovery then switches to the registered implementation, causing nondeterminism, step conflicts, or incorrect results.

maxdml:
Thanks @ibrhmkoz . `RunWorkflow` silently executing a different function you registered is indeed a serious issue. Note that, if you were to call `RegisterWorkflow` on both function values, you'd have a panic because we'd have detected the collision. I'll put on a fix that have `RunWorkflow` run the registered function, no matter what, and document better what can be registered and what not. For the next major release, we might be introducing concrete `Workflow` object that will resolve this problem.

ibrhmkoz:
Yes, instead of RunWorkflow, creating a new self contained Workflow precludes such bug via SDK design.

PR #347, opened 2026-06-10
Support configured instance + fix collision execution

maxdml:
#345 exposed a subtle execution bug and a workflow registration limitation. Specifically, Go functions can share the same function entry and return the same name, for example if they are receiver methods. This 1) limits too stringently what can be registered and 2) allows direct execution of non-registered workflows.

This PR does two things:
- Always use the registered workflow during direct execution
- Introduce configured instances

Users can register receiver methods:

```go
dbos.RegisterWorkflow(ctx, slack.Send, dbos.WithInstance(slack))
dbos.RegisterWorkflow(ctx, email.Send, dbos.WithInstance(email))
```

And call them as:

```go
dbos.RunWorkflow(ctx, slack.Send, input, dbos.WithRunInstance(slack))
```

A configured instance must implement the interface:

```go
type ConfiguredInstance interface {
	ConfigName() string
}
```

Note that this model doesn't support registering a variety of closures, non-inlined methods, etc

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -633,10 +633,25 @@ func resolveWorkflowFunctionName[P any, R any](fn Workflow[P, R]) string {
 // RegisterWorkflow registers a function as a durable workflow that can be executed and recovered.
 // The function is registered with type safety - P represents the input type and R the return type.
 //
+// Workflows are identified by a name derived from the function's code pointer, so each
+// registered function value must have a unique name. Registrable:
+//   - Top-level named functions: the recommended form. Each has a unique name.
+//   - Generic function instantiations: type parameters are automatically appended to the name,
+//     so distinct instantiations are distinct workflows.
+//   - Method values bound to a configured instance (e.g. inst.Run), registered with
+//     WithInstance: the instance's config name qualifies the workflow name, so each
+//     instance registers its own workflow. Run these with WithRunInstance.
+//   - A closure or method value, at most ONE per source expression: all values built
+//     from the same func literal or method (e.g. a.Run and b.Run, or closures from one
+//     factory) share a name. Registering a second one panics with
+//     ConflictingRegistrationError; use WithInstance (methods) or distinct top-level
+//     functions (closures) instead.
+//
 // Registration options include:
 //   - WithMaxRetries: Set maximum retry attempts for workflow recovery
 //   - WithSchedule: Register as a scheduled workflow with cron syntax
-//   - WithWorkflowName:: Set a custom name for the workflow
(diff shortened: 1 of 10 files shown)


ibrhmkoz:
Instead, we can create workflows as follows:

```go
// Workflow is a callable, self-contained, durable workflow.
type Workflow[P any, R any] func(ctx DBOSContext, input P, opts ...WorkflowOption) (*WorkflowHandle[R], error)

// NewWorkflow creates and registers a workflow.
// Calling the returned function persists an execution request for an available worker.
// Execution behavior is configured with options such as WithGlobalConcurrency and WithRateLimit.
func NewWorkflow[P any, R any](ctx DBOSContext, fn WorkflowFn[P, R], opts ...WorkflowOption) Workflow[P, R] {
```

and call them as follows:

```go
simpleWorkflowD := func simpleWorkflow(dbosCtx DBOSContext, input string) (string, error) {
	return input, nil
}
simpleWorkflowD := NewWorkflow(dbosCtx, simpleChildWf)
handle, err := simpleWorkflowD(dbosCtx, input, opts...)
```

That would prevent this misbehavior and create a more concise abstraction that allows us to call a workflow like a normal function.

maxdml:
@ibrhmkoz maybe we'll introduce this in v1 of the library

dbos/workflow.go
```diff
@@ -580,6 +597,20 @@ func WithWorkflowName(name string) WorkflowRegistrationOption {
...
 
+// WithInstance registers a workflow method bound to a specific configured instance.
+// Method values bound to different receivers (e.g. a.Run and b.Run) share a function
+// name, so each instance's method must be registered under a per-instance key:
+//
+//	dbos.RegisterWorkflow(ctx, slack.Send, dbos.WithInstance(slack))
+//	dbos.RegisterWorkflow(ctx, email.Send, dbos.WithInstance(email))
+//
+// Run the workflow with the matching dbos.WithRunInstance option.
+func WithInstance(instance ConfiguredInstance) WorkflowRegistrationOption {
```
kraftp, line 608: How does this interact with things that look up workflows, like schedules or the debouncer?

maxdml: Good catch, I must add support for these. Specifically:
- Calls with direct functions must pass the actual instance (on which we can call `ConfigName()`)
- Calls from the client must pass the config name manually (enqueue, debouncer, create schedule, apply schedules)

I'm adding this for the debouncer, however, for scheduled workflows, we'll need a migration to add the config name in the schedule, because the code reconstruct the name for a registry look-up when reconciling schedules.

kraftp (approved):

PR #352, opened 2026-06-15
Enable Conductor for Cloud

maxdml:
Also fix a small race in cancellation status handling


kraftp (commented):
This looks good--can we do a quick manual test Cloud side?

kraftp (approved):

PR #354, opened 2026-06-16
DB-backed queues

maxdml:
+ fix a couple places where we called `retry`/`retryWithResult` with an uncancellable context.

### DIFF
--- a/dbos/queue.go
+++ b/dbos/queue.go
@@ -338,6 +338,14 @@ type queueRunner struct {
 	listenMu       sync.Mutex
 	listenedQueues map[string]bool
 
+	// currentQueues holds the latest reconciled set of queues this process runs
+	// workers for (the in-memory registry plus database-backed queues, filtered by
+	// the listen set). The supervisor rebuilds it once per reconcile tick by
+	// replacing the reference, never mutating in place; workers read their
+	// configuration from it.
+	currentMu     sync.RWMutex
+	currentQueues map[string]WorkflowQueue
+
 	// WaitGroup to track all queue goroutines
 	queueGoroutinesWg sync.WaitGroup
 
@@ -353,6 +361,7 @@ func newQueueRunner(logger *slog.Logger) *queueRunner {
 		jitterMax:             1.05,
 		workflowQueueRegistry: make(map[string]WorkflowQueue),
 		listenedQueues:        make(map[string]bool),
+		currentQueues:         make(map[string]WorkflowQueue),
 		completionChan:        make(chan struct{}, 1),
 		logger:                logger.With("service", "queue_runner"),
 	}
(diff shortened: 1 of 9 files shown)


dbos/queue.go
```diff
@@ -214,51 +370,135 @@ func (qr *queueRunner) getQueue(queueName string) *WorkflowQueue {
...
+	workerDone := make(map[string]chan struct{})
+
+	const reconcileInterval = 1 * time.Second
+	for ctx.Err() == nil { // While ctx is not cancelled
+		// Transition any DELAYED workflows whose delay has expired to ENQUEUED.
+		if err := retry(ctx, func() error {
+			return ctx.systemDB.transitionDelayedWorkflows(ctx)
+		}, withRetrierLogger(qr.logger)); err != nil {
+			qr.logger.Warn("Exception transitioning delayed workflows", "error", err)
+		}
```
maxdml, line 554: Do this in the queue runner management loop (instead of doing it in each queue goroutine...)

dbos/queue.go
```diff
@@ -214,51 +370,135 @@ func (qr *queueRunner) getQueue(queueName string) *WorkflowQueue {
...
+			}
+			current[queue.Name] = queue
+		}
 	}
 
-	for _, queue := range queuesToListenMap {
-		qr.queueGoroutinesWg.Add(1)
-		go qr.runQueue(ctx, queue)
+	if internal, ok := qr.workflowQueueRegistry[_DBOS_INTERNAL_QUEUE_NAME]; ok {
+		current[_DBOS_INTERNAL_QUEUE_NAME] = internal
```
maxdml: internal queue is still an in-memory queue

dbos/queues_test.go
maxdml, line 1: Moved all the queue test to DB-backed queues, + added a few tests for CRUD/guardrails and mixed queues (inmemory + db). There's still a few in-memory queues in the other suites.

dbos/system_database.go
```diff
@@ -4274,6 +4280,161 @@ func (s *sysDB) getQueuePartitions(ctx context.Context, queueName string) ([]str
...
+	if err := row.Scan(&name, &concurrency, &workerConcurrency, &rateLimitMax, &rateLimitPeriodSec, &priorityEnabled, &partitionQueue, &pollingIntervalSec); err != nil {
+		return nil, err
+	}
+	q := &WorkflowQueue{
+		Name:                 name,
+		GlobalConcurrency:    concurrency,
+		WorkerConcurrency:    workerConcurrency,
+		PriorityEnabled:      priorityEnabled,
+		PartitionQueue:       partitionQueue,
+		MaxTasksPerIteration: _DEFAULT_MAX_TASKS_PER_ITERATION, // not persisted; queue table has no such column
```
maxdml, line 4315: imo max tasks per iteration and max polling interval should be in the schema.

kraftp: MaxTasksPerIteration should be purged completely, I think. Performance tests showed it was unhelpful.

dbos/workflow.go
```diff
@@ -1197,6 +1197,20 @@ func (c *dbosContext) RunWorkflow(_ DBOSContext, fn WorkflowFunc, input any, opt
 	// Validate queue exists if provided
 	if len(params.QueueName) > 0 {
 		queue := c.queueRunner.getQueue(params.QueueName)
+		if queue == nil {
```
kraftp: We dropped this in Python/TS to eliminate extra roundtrips

dbos/system_database.go
```diff
@@ -90,6 +90,12 @@ type systemDatabase interface {
 	clearQueueAssignment(ctx context.Context, workflowID string) (bool, error)
 	getQueuePartitions(ctx context.Context, queueName string) ([]string, error)
 
+	// Database-backed queue registry (the queues table)
```
kraftp, line 93: I don't see dynamic setters in this PR?

dbos/system_database.go
maxdml: I can remove it from the public interface in v1

kraftp (approved):

PR #358, opened 2026-06-18
Datasources

### LINKED ISSUE #336: Datasources/Transactional Steps
These are steadily increasing in popularity in the other three languages. We should design an good interface for them in Go.

maxdml:
# Usage

```go
pool, _ := pgxpool.New(ctx, appDatabaseURL)
ds, err := dbos.NewDataSource(dbosCtx, pool, dbos.WithDataSourceName("app"))
// Options: WithDataSourceName (logs only), WithDataSourceSchema (default "dbos").
// NewDataSource is dynamic — no registry. Call it any time, before or after Launch.

// Inside a workflow:
n, err := dbos.RunAsTransaction(ctx, ds, func(ctx context.Context, tx dbos.Tx) (int64, error) {
    res, err := tx.Exec(ctx, "INSERT INTO orders(item) VALUES ($1)", item)
    if err != nil {
        return 0, err
    }
    return res.RowsAffected()
}, dbos.WithStepMaxRetries(3)) // standard StepOptions apply
```

Must be called **from within a workflow**. Serialization/deadlock conflicts are retried internally with a fresh tx; application errors follow your step retry policy

# Nesting
Running inside another `RunAsTransaction` is rejected. Running inside a `RunAsStep` is allowed but results in at-least-once guarantees. Exactly what you'd have managing your own transaction inside a DBOS step. (We can also decided to outright reject this 🤷. Not a big deal either way.)

# Permissions
`NewDataSource` eagerly provisions the durability table in your database and will return you an error if the table needs to be created, but the engine you passed doesn't have permissions.

# Same-engine optimization

If the data source's pool is the **same engine handle** as the DBOS system database (you passed your pool as `Config.SystemDBPool` / `Config.SqliteSystemDB` and reuse it here), `RunAsTransaction` collapses onto the single system transaction: app writes and the `operation_outputs` checkpoint commit **together**, no `transaction_completion` table is created, and recovery replays from `operation_outputs` alone. Detection is by **pointer identity** (not DSN)

Closes #336

### DIFF
--- a/dbos/client_test.go
+++ b/dbos/client_test.go
@@ -1287,6 +1287,18 @@ func TestListWorkflows(t *testing.T) {
 			assert.Equal(t, expectedOffsetThree[i].ID, wf.ID, "offset workflow at index %d: unexpected ID", i)
 		}
 
+		// Offset without a limit: SQLite rejects a bare OFFSET, so this exercises
+		// the dialect's "no limit" sentinel. Expect all workflows after the first 5.
+		offsetNoLimitWorkflows, err := client.ListWorkflows(
+			WithWorkflowIDPrefix("test-"),
+			WithOffset(5))
+		require.NoError(t, err, "failed to list workflows with offset and no limit")
+		expectedOffsetNoLimit := ascWorkflows[5:]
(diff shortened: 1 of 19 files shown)


kraftp (approved):
This looks correct as far as I can tell!

PR #364, opened 2026-06-29
fix(stream): drain to the end of the stream when the producer goes inactive

SamuelXing:
A stream reader can drop the producer's last value if it's written right as the producer finishes.

`readStream` reads the stream, then makes a separate call to check whether the producing workflow is still active. If the producer commits its final value and completes in the gap between those two calls, the reader sees the workflow is done and closes the stream without ever reading that value.

The window is small, and it only affects a producer that finishes *without* calling `CloseStream` (so the reader is relying on the "stop once the workflow goes inactive" fallback) while another reader is draining concurrently. The normal `CloseStream` path is fine — the close sentinel guarantees the reader sees everything.

**Fix:** once the reader notices the producer has gone inactive, do one more read before closing. All of a workflow's writes are committed by the time its status is terminal, so that final pass reliably catches anything written just before it finished.

**This matches Python.** The Python SDK fixed the same race in [dbos-inc/dbos-transact-py#713](https://github.com/dbos-inc/dbos-transact-py/pull/713): its `read_stream` keeps a `final_read` flag so that when it sees the workflow has gone inactive it reads to the end of the stream before stopping, instead of returning right away. 

**Test:** a deterministic unit test (no database needed) that forces the exact interleaving with a fake `systemDatabase` — first read empty, producer reported terminal, value shows up on the next read. It fails without the fix and passes with it.

### DIFF
--- a/dbos/read_stream_drain_test.go
+++ b/dbos/read_stream_drain_test.go
@@ -1,0 +1,59 @@
+package dbos
+
+import (
+	"context"
+	"io"
+	"log/slog"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// fakeStreamDB implements only the two systemDatabase methods that readStream
+// uses. It embeds the interface so it satisfies the rest (none of which are
+// called here) without a pile of stub methods.
+type fakeStreamDB struct {
+	systemDatabase
+	reads int
+}
+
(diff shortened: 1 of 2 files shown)


SamuelXing:
hi @maxdml, can I have you review on this one?

maxdml (approved):

maxdml:
Thanks @SamuelXing !

PR #369, opened 2026-07-02
Draft: transaction step error-rollback is inconsistent across data sources

SamuelXing:
Draft — the test here is meant to fail, I'm not asking to merge it. I noticed transaction steps behave differently on error depending on the data source, and I wanted to check whether that's intended before changing anything.

If a transaction body writes a row and then returns an error:

- on a separate data source, the write rolls back — that's what `TestRunAsTransaction/RollsBackOnError` checks
- on a data source that shares the system DB pool, the call goes through `runAsTxn`, which records the error inside the same transaction and commits it — so the row stays

The test runs the same failing transaction on both and prints what each leaves behind:

```
rows left by the failed transaction — separate data source: 0, shared system-DB pool: 1
```

So the same user code either rolls back or leaves a partial write, depending only on whether the data source happens to share the system pool. Both paths came in with #358.

### To reproduce

```sh
git fetch origin test/runastxn-partial-write-on-error
git checkout test/runastxn-partial-write-on-error

# SQLite — no database needed:
DBOS_TEST_BACKEND=sqlite go test ./dbos/ -run TestTransactionErrorRollbackIsInconsistent -v

# Postgres (uses the SDK's default local DB URL):
DBOS_SYSTEM_DATABASE_URL=postgres://postgres:dbos@localhost:5432/dbos?sslmode=disable \
  go test ./dbos/ -run TestTransactionErrorRollbackIsInconsistent -v
```

The `FAIL` is expected — the line to read is `separate data source: 0, shared system-DB pool: 1`. Same result on both backends.

### DIFF
--- a/dbos/transaction_error_rollback_test.go
+++ b/dbos/transaction_error_rollback_test.go
@@ -1,0 +1,72 @@
+package dbos
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+// TestTransactionErrorRollbackIsInconsistent runs the SAME failing transaction
+// step against two data sources and shows they disagree on whether the body's
+// write is rolled back.
+//
+// A transaction body writes a row and then returns an application error:
+//   - on a separate data source, the write is rolled back (row count 0)
+//   - on a data source that shares the system-DB pool (the runAsTxn path), the
+//     write is committed along with the error record (row count 1)
+//
+// The second is the bug this documents. The test asserts both roll back, so it
+// FAILS on the shared-pool path — intentionally. See the PR description.
+func TestTransactionErrorRollbackIsInconsistent(t *testing.T) {
+	// A shared data source (pool == system DB) and a separate one, one context.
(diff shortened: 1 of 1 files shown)


SamuelXing:
Hi @maxdml 

I'm not sure this is a bug. Committing the write and the error record in one transaction means a crash can't leave them out of sync — the exact gap the separate path needs its `checkCompletion` step to cover. So I'd rather ask than just flip it.

For what it's worth, both the separate path and [the Python SDK](https://github.com/dbos-inc/dbos-transact-py/blob/f6cfec5dc0a4318a6569fc789bf94d6bf3c73dce/dbos/_core.py#L1463-L1581) roll back on error (Python re-raises out of `with session.begin()` and records the failure separately), so there's precedent for rolling back.

If you want the shared path to do the same, I can push the fix on this branch (roll back on error, record the failure separately, add the completion check) and this test goes green. Otherwise a short comment would help, since it's the one path that keeps the partial write.

maxdml:
Thanks @SamuelXing . Great catch. Yeah, the shared database path should rollback the user changes when executing a user transaction (while checkpointing the error durably.) I'll fix it ASAP

PR #370, opened 2026-07-02
Fix GC row-threshold ties on created_at

amrrdev:
## Summary

This fixes flaky garbage-collection behavior when multiple workflows share the same `created_at` millisecond.

The row-threshold GC path selected the Nth newest workflow using only `created_at`, then deleted workflows with `created_at < cutoff`. When several workflows were created within the same millisecond as the cutoff row, they were all retained, causing GC to keep more rows than the configured threshold.

## Root Cause

`created_at` is stored with millisecond precision, so it is not guaranteed to be unique. The existing logic treated the cutoff as a timestamp, while the threshold query was actually selecting a specific row. That mismatch caused the flaky behavior.

## Fix

The row-threshold cutoff is now deterministic by using `workflow_uuid` as a tiebreaker.

The cutoff row is selected with:

```sql
ORDER BY created_at DESC, workflow_uuid DESC
```

This ensures GC retains exactly the intended number of newest workflows, even when multiple workflows share the same `created_at` timestamp.

## Verification

Ran:

```bash
DBOS_TEST_BACKEND=sqlite go test ./dbos -run '^TestGarbageCollect/(GarbageCollectWithOffset|GarbageCollectOnlyCompletedWorkflows)$' -count=100

go test ./dbos -run '^TestGarbageCollect$'
```

### DIFF
--- a/dbos/system_database.go
+++ b/dbos/system_database.go
@@ -1786,16 +1786,19 @@ func (s *sysDB) garbageCollectWorkflows(ctx context.Context, input garbageCollec
 	}
 
 	cutoffTimestamp := input.cutoffEpochTimestampMs
+	var rowsBasedCutoffWorkflowID string
+	useRowsThresholdCutoff := false
 
-	// If rowsThreshold is provided, get the timestamp of the Nth newest workflow
+	// If rowsThreshold is provided, get the Nth newest workflow using workflow_uuid
+	// as a deterministic tiebreaker for workflows created in the same millisecond.
 	if input.rowsThreshold != nil {
-		query := s.renderSQL(`SELECT created_at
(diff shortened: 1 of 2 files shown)


maxdml:
Thanks @amrrdev . I thought about this a little bit. The challenge with this PR approach is that adding a new column to the SELECT makes the query much slower. There's currently an index on `created_at`, so the select query can be satisfied by the index only, whereas querying `created_at`, `workflow_uuid` forces heap lookups.

`workflow_status` being in the hot path, and the collision fairly rare, I think we can just ignore this and make the test more robust. Not a big deal. If we really have requests for exact retention counts, we can do a 2nd delete pass after the first delete query (that'd be much faster)

PR #375, opened 2026-07-03
Include was_forked_from in workflow export/import

SamuelXing:
Forking a workflow sets `was_forked_from = true` on the original (the workflow the fork was taken from). It's a real column: the fork writes it, `ListWorkflows` reads it back, and `WithWasForkedFrom(true)` filters on it.

Export/import doesn't preserve it. If you export a workflow that's been forked from and import it elsewhere (or back into an empty DB), `was_forked_from` comes back `false`.

It's just an omission in the column lists. Both the export SELECT and the import INSERT spell out their columns by hand, and they copy `forked_from` but never listed `was_forked_from`. So on import the column falls back to its `NOT NULL DEFAULT FALSE` and the flag is gone. Nothing errors, it just flips.

Python already handles this: it selects `was_forked_from` on export, carries it over on import (defaulting to `False` for older payloads), and has a test for the round trip. Pinned to `dbos-inc/dbos-transact-py@f6cfec5`:

- export, selects the column: [`_sys_db.py#L4501`](https://github.com/dbos-inc/dbos-transact-py/blob/f6cfec5dc0a4318a6569fc789bf94d6bf3c73dce/dbos/_sys_db.py#L4501), puts it in the payload: [`_sys_db.py#L4545`](https://github.com/dbos-inc/dbos-transact-py/blob/f6cfec5dc0a4318a6569fc789bf94d6bf3c73dce/dbos/_sys_db.py#L4545)
- import, carries it over: [`_sys_db.py#L4706`](https://github.com/dbos-inc/dbos-transact-py/blob/f6cfec5dc0a4318a6569fc789bf94d6bf3c73dce/dbos/_sys_db.py#L4706)
- test, asserts the round trip: [`test_workflow_export.py#L93`](https://github.com/dbos-inc/dbos-transact-py/blob/f6cfec5dc0a4318a6569fc789bf94d6bf3c73dce/tests/test_workflow_export.py#L93)

So right now Go and Python disagree on what an export/import keeps. This lines Go up with Python.

## The fix

- Export: add `was_forked_from` to the `workflow_status` SELECT, scan it, and put it in the payload.
- Import: add it to the INSERT. The column is `NOT NULL`, so a payload from before this change won't have the key; default it to `false` (same as Python's `.get("was_forked_from", False)`) so old exports still import cleanly.

## How to reproduce

Run the test in this PR against `main`:

```bash
DBOS_TEST_BACKEND=sqlite go test ./dbos/ -run TestExportImportPreservesWasForkedFrom -v
```

On `main` it fails with `was_forked_from must be present in the export payload`; with this change it passes. Same on Postgres.

By hand it's: run a workflow, fork it (that marks the original), export the original, delete it, re-import, and check `WasForkedFrom` on the reimported row. It's `false` before the fix and `true` after.

## Testing

Added `TestExportImportPreservesWasForkedFrom`: run a workflow, fork it, export the source, delete it, re-import, and assert `was_forked_from` is still `true` (and that `WithWasForkedFrom(true)` still finds it). It fails without the fix and passes with it, on both SQLite and Postgres. `TestExportImportWorkflow` still passes, and `go vet` / `gofmt` are clean.

### DIFF
--- a/dbos/workflows_test.go
+++ b/dbos/workflows_test.go
@@ -7370,80 +7370,6 @@ func TestExportImportWorkflow(t *testing.T) {
 	})
 }
 
-// TestExportImportPreservesWasForkedFrom checks that the was_forked_from flag —
-// which marks a workflow a fork was taken *from* (the fork source), not the fork
-// itself — survives an export/import round trip.
-//
-// Regression: the export/import status column list included forked_from but not
-// was_forked_from, so re-importing a fork source silently lost the flag (it
-// reverted to the column's NOT NULL DEFAULT FALSE).
-func TestExportImportPreservesWasForkedFrom(t *testing.T) {
-	dbosCtx := setupDBOS(t, setupDBOSOptions{dropDB: true, checkLeaks: true})
-	RegisterWorkflow(dbosCtx, simpleWorkflow)
-	Launch(dbosCtx)
-
-	sdb := dbosCtx.(*dbosContext).systemDB.(*sysDB)
-
-	lookup := func(ids ...string) map[string]WorkflowStatus {
-		wfs, err := sdb.listWorkflows(dbosCtx, listWorkflowsDBInput{workflowIDs: ids})
-		require.NoError(t, err)
-		m := make(map[string]WorkflowStatus, len(wfs))
-		for _, wf := range wfs {
-			m[wf.ID] = wf
-		}
-		return m
-	}
-
-	// Run a workflow, then fork it. Forking marks the ORIGINAL was_forked_from.
-	sourceID := uuid.NewString()
-	handle, err := RunWorkflow(dbosCtx, simpleWorkflow, "hello", WithWorkflowID(sourceID))
-	require.NoError(t, err)
-	_, err = handle.GetResult()
-	require.NoError(t, err)
-
-	forkHandle, err := ForkWorkflow[string](dbosCtx, ForkWorkflowInput{OriginalWorkflowID: sourceID})
-	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


maxdml:
Thanks @SamuelXing . LGTM. Can you just merge this test with one of the existing fork test? It's just about checking a few fields -- I don't want to add another test just for it

SamuelXing:
> Thanks @SamuelXing . LGTM. Can you just merge this test with one of the existing fork test? It's just about checking a few fields -- I don't want to add another test just for it

sure, just updated, could you re-review?

dbos/workflows_test.go
```diff
@@ -7368,6 +7368,54 @@ func TestExportImportWorkflow(t *testing.T) {
...
+		require.NoError(t, sdb.deleteWorkflows(dbosCtx, deleteWorkflowsDBInput{
+			workflowIDs: []string{srcID, forkID},
+		}))
+		require.NoError(t, sdb.importWorkflow(dbosCtx, exported))
+
+		roundTripped, err := sdb.listWorkflows(dbosCtx, listWorkflowsDBInput{workflowIDs: []string{srcID}})
+		require.NoError(t, err)
+		require.Len(t, roundTripped, 1)
+		assert.True(t, roundTripped[0].WasForkedFrom,
+			"was_forked_from must survive an export/import round trip")
```
maxdml: What I meant is to add this single test assertion into an existing test (e.g., `TestFork`)

maxdml (approved):

maxdml:
Thanks @SamuelXing

PR #384, opened 2026-07-09
Fix step cancellation outcome

maxdml:
- Trigger a durable cancel when the context is manually cancelled, too. Also forbid updating workflow outcome when status is cancelled.
- Stop recording durably cancellation errors for steps and transactions
- Stop updating workflow outcomes when the status is already CANCELLED.
- Fix a race between clearing the activeWorkflowIDs map and the workflow outcome being writen (the latter happens before the former and can lead to stuck workflows)

Also add a new chaos suite for context cancellation, checking that the invariant "cancelled workflow can be resumed".

### DIFF
--- a/dbos/client_test.go
+++ b/dbos/client_test.go
@@ -1096,10 +1096,19 @@ func TestListWorkflows(t *testing.T) {
 		// Record start time for filtering tests
 		testStartTime := time.Now()
 
+		// Boundary between the test-batch-* and test-other-* workflows, observed
+		// rather than derived from the sleep cadence below: enqueue latency (notably
+		// on CockroachDB) makes wall-clock arithmetic unreliable.
+		var firstHalfTime time.Time
+
 		// Start 10 workflows at 100ms intervals with different patterns
 		for i := range 10 {
 			var workflowID string
 			var handle WorkflowHandle[string]
+
+			if i == 5 {
+				firstHalfTime = time.Now()
+			}
 
 			if i < 5 {
 				// First 5 workflows: use prefix "test-batch-" and succeed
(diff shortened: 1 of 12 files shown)


dbos/client_test.go
maxdml, line 1: Fix a flaky test

dbos/workflow.go
```diff
@@ -244,15 +248,18 @@ func checkGetResultExecution[R any](dbosCtx context.Context) (R, bool, error) {
...
-			return *new(R), false, fmt.Errorf("failed to resolve decoder: %w", err)
-		}
-		decodedOutput, err := decoder.Decode(recordedOutputs.output)
-		if err != nil {
-			return *new(R), false, fmt.Errorf("failed to decode operation result: %w", err)
+		var decodedOutput R
+		if recordedOutputs.output != nil {
+			decoder, err := resolveDecoder[R](recordedOutputs.serialization, dbosCtx.(*dbosContext).serializer)
+			if err != nil {
+				return *new(R), false, fmt.Errorf("failed to resolve decoder: %w", err)
```
kraftp, line 255: If I understand this correctly, if a step returns (non-zero, err) it will now always return (zero, err)?

dbos/workflow.go
```diff
@@ -1596,32 +1643,45 @@ func (c *dbosContext) RunWorkflow(_ DBOSContext, fn WorkflowFunc, input any, opt
 				ser = awaitOut.serialization
 			}
 			// Keep the encoded result - decoding will happen in RunWorkflow[P,R] when we know the target type
-			outcomeChan <- workflowOutcome[any]{result: encodedResult, err: err, needsDecoding: true, serialization: ser}
+			outcomeChan <- workflowOutcome[any]{result: encodedResult, err: err, needsDecoding: true, serialization: ser,
+				cancelled: errors.Is(err, &DBOSError{Code: AwaitedWorkflowCancelled})}
```
kraftp, line 1653: Should we call `stopFunc` here?

kraftp (approved):

dbos/workflow.go
maxdml: It'll only return `nil, err` if there was a decoding error. Otherwise, it'll return `non-nil, err`

dbos/workflow.go
maxdml: good catch, we should

PR #390, opened 2026-07-14
Subpackages

maxdml:
In preparation of the "db driver registration feature" (so that pg users don't have to import sqlite deps), move much of the internal logic into formal `internal` subpackages.

- All data structures are in `internal/model`
- sysdb in `internal/sysdb`

All of these changes are mechanical. The diff in the tests is just for adding package prefixes, etc.

I'm not moving the tests into subpackages (they import them) to keep the PR mechanical & simple.


dbos/dbos.go
```diff
@@ -616,20 +620,28 @@ func NewDBOSContext(ctx context.Context, inputConfig Config) (DBOSContext, error
...
+		Logger:          initExecutor.logger,
+		ApplicationName: config.AppName,
+		EncodeScheduledInput: func(ctx context.Context, scheduledTime time.Time, scheduleContext any) (*string, string, error) {
+			ser := resolveEncoder(ctx)
+			encoded, err := ser.Encode(ScheduledWorkflowInput{
+				ScheduledTime: scheduledTime,
+				Context:       scheduleContext,
+			})
+			return encoded, ser.Name(), err
+		},
```
maxdml, line 636: Pass a closure to the sysdb so we don't leak the context

dbos/aliases.go
maxdml, line 1: Keep public types public

kraftp (approved):

PR #392, opened 2026-07-14
Distinguish between our own checkpointing writes and conflicting writes

maxdml:
Checkpointing a step and recording a child workflow in the operations_outcome table can return a conflict error if the operation is retried automatically as part of of DB transient errors handling. Specifically, if a connection errors happen between the commit and its acknowledgment by the client, we'd naïvely retry and return an error upon conflict.

This PR handles with this:
- Add a `ON CONFLICT DO NOTHING` statement
- When no rows were affected, check whether re-reading the row yields inconsistent results

note: in the step checkpointing case, we do return a conflict error if the value is not found anymore. I don't think this should happen often tho. (It'd require another concurrent delete of the same conflicting row ...)

Also:
- Fix: do not set deadlines during enqueue

### DIFF
--- a/dbos/internal/sysdb/system_database.go
+++ b/dbos/internal/sysdb/system_database.go
@@ -2594,8 +2594,7 @@ func (s *SysDB) RecordOperationResult(ctx context.Context, input RecordOperation
 		&storedOutput, &storedError, &storedFunctionName, &storedSerialization, &storedChildID, &storedStartedAtMs, &storedCompletedAtMs)
 	if err != nil {
 		if err == pgx.ErrNoRows {
-			// The conflicting row is not visible to this transaction's snapshot
-			// It means a concurrent writer inserted the row
+			// This should only happen if the conflicting row was deleted, e.g., during GC
 			return models.NewWorkflowConflictIDError(input.WorkflowID)
 		}
 		return fmt.Errorf("failed to read existing operation result: %w", err)
--- a/dbos/internal/sysdb/system_database.go
+++ b/dbos/internal/sysdb/system_database.go
@@ -2530,6 +2530,19 @@ type RecordOperationResultDBInput struct {
 	Serialization   string
 }
 
+// RecordOperationResult checkpoints a step outcome. A checkpoint already
+// existing at (workflow_uuid, function_id) is disambiguated by content:
+//   - identical to input (including the caller's timestamps) → our own earlier
+//     write whose commit ack was lost; the retry is a no-op success.
+//   - different function name → determinism violation (UnexpectedStep).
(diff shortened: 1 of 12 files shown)


dbos/internal/sysdb/system_database.go
```diff
@@ -2546,24 +2559,75 @@ func (s *SysDB) RecordOperationResult(ctx context.Context, input RecordOperation
...
+	if err != nil {
+		if err == pgx.ErrNoRows {
+			// This should only happen if the conflicting row was deleted, e.g., during GC
 			return models.NewWorkflowConflictIDError(input.WorkflowID)
 		}
-		return err
+		return fmt.Errorf("failed to read existing operation result: %w", err)
+	}
+	if input.StepName != storedFunctionName {
+		return models.NewUnexpectedStepError(input.WorkflowID, input.StepID, input.StepName, storedFunctionName)
```
kraftp: Notably, this check is now before the check for NewWorkflowConflictIDError, which I don't think is right

dbos/internal/sysdb/system_database.go
```diff
@@ -2579,35 +2643,47 @@ type RecordChildWorkflowDBInput struct {
...
+		var recordedChildID *string
+		if err := querier.QueryRow(ctx, selectQuery, input.ParentWorkflowID, input.StepID).Scan(&recordedChildID); err != nil {
+			return fmt.Errorf("failed to check existing child workflow record: %w", err)
+		}
+		if recordedChildID == nil || *recordedChildID != input.ChildWorkflowID {
+			recorded := "<nil>"
+			if recordedChildID != nil {
+				recorded = *recordedChildID
+			}
+			return fmt.Errorf(
```
kraftp: Should this be models.NewWorkflowConflictIDError?

kraftp (approved):

dbos/internal/sysdb/system_database.go
maxdml: Ha, yes, the sequence should be:
1. Check if this is our write
2. If not, is this a non determinism error?
3. Else, another execution of the same workflow is running, park the current execution

maxdml: Fixed. That being said, I actually don't think the order mattered. A concurrent execution we want to park for would never have a different function name.

dbos/internal/sysdb/system_database.go
maxdml: Maybe `NewUnexpectedStepError`. `NewWorkflowConflictIDError` is what we use to decide whether the workflow must be parked and return a polling handle. It is used when we detect a concurrent execution raced ahead of the local one.

PR #401, opened 2026-07-17
Remove deprecated in-memory queue/schedules

maxdml:
Also:
- WithQueue(queue Queue) — takes a queue object
- ForkWorkflow[R](ctx, id, opts...) — new ForkOptions: WithForkStartStep/WorkflowID/ApplicationVersion/Queue/QueuePartitionKey
- ListRegisteredWorkflows — error return dropped
- WithTxIsolation(IsoLevel) — new StepOption, no-op outside transactions
- ListenQueues(ctx, names ...string) — variadic names
- DequeueWorkflows — removed max task per iteration (already ignored for db backed queues)

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,9 +152,12 @@ func main() {
         panic(err)
     }
 
-    // Register the workflow and create a durable queue
+    // Register the workflow and a durable queue
     dbos.RegisterWorkflow(ctx, task)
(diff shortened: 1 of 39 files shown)


kraftp (approved):
Yeah, good idea to clean these up in V1

PR #404, opened 2026-07-18
v1.0.0-rc.1 p6

maxdml:
Lots of polish:
- Enqueue[P, R] becomes Enqueue[R, P]; input type inferred
- WithAuthenticatedRoles/WithEnqueueAuthenticatedRoles now variadic, not slice
- WorkflowQueue struct unexported as workflowQueue; Queue interface stays
- GobSerializer struct unexported; constructor NewGobSerializer kept
- ClearRegistries unexported (test-only helper)
- ErrorCodeConflictingWorkflow renamed ErrorCodeUnexpectedWorkflow, new ErrUnexpectedWorkflow sentinel
- New CauseKind preserves context.Canceled/DeadlineExceeded across DB round-trips
- Error gets stable JSON marshaling: message plus symbolic code
- WorkflowStatus JSON: timeout_ms milliseconds, omitzero timestamps, custom unmarshal
- Listing paths decode payloads via serializer, not raw strings
- Schedule Context becomes json.RawMessage; new DecodeScheduleContext[T] helper
- GetSchedule/GetLatestApplicationVersion return values instead of pointers
- Launch now recovers pending workflows before starting queue runner
- Recovery skips dead-lettered workflows instead of aborting entirely
- `ListenQueues` now replaces whole set; new `ListenedQueues` accessor. Now that the in-memory queue are gone, `ListenQueue` can be made more flexible.
- Shutdown reports components that missed the timeout (sysdb, conductor)
- Enqueue insert wrapped in retry on retryable transaction errors
- Schedule/queue DB writes gain IsRetryableTransaction retry condition
- Documented Serializer nil-handling contract and __DBOS_NIL sentinel
- New plainError keeps legacy error strings readable in JSON
- Integration mocks moved to integration/internal/mocks

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -708,9 +708,9 @@ func NewClient(ctx context.Context, config ClientConfig) (Client, error) {
 	return dbosCtx, nil
 }
 
-// Launch initializes and starts the DBOS runtime components including the system database,
-// admin server (if enabled), queue runner, workflow scheduler, and performs recovery
-// of any pending workflows on this executor.
+// Launch initializes and starts the DBOS runtime components including the system database
+// and admin server (if enabled), recovers any pending workflows on this executor, then
+// starts the queue runner and workflow scheduler.
 //
 // Returns an error if the context is already launched or if any component fails to start.
 func (c *dbosContext) Launch() error {
@@ -754,6 +754,18 @@ func (c *dbosContext) Launch() error {
 		c.logger.Debug("Admin server started", "port", c.config.AdminServerPort)
 	}
 
+	// Recover local pending workflows before starting the queue runner so
+	// recovered workflows are not racing a fresh dequeue pass.
+	recoveryHandles, err := recoverPendingWorkflows(c, []string{c.executorID})
+	if err != nil {
+		return models.NewInitializationError(fmt.Sprintf("failed to recover pending workflows during launch: %v", err))
+	}
+	if len(recoveryHandles) > 0 {
+		c.logger.Info("Recovered pending workflows", "count", len(recoveryHandles))
+	} else {
+		c.logger.Debug("No pending workflows to recover")
+	}
+
 	// Start the queue runner in a goroutine
 	go func() {
 		c.queueRunner.run(c)
@@ -778,17 +790,6 @@ func (c *dbosContext) Launch() error {
 	if c.conductor != nil {
 		c.conductor.launch()
 		c.logger.Debug("Conductor started")
-	}
-
-	// Run a round of recovery on the local executor
-	recoveryHandles, err := recoverPendingWorkflows(c, []string{c.executorID})
-	if err != nil {
-		return models.NewInitializationError(fmt.Sprintf("failed to recover pending workflows during launch: %v", err))
-	}
-	if len(recoveryHandles) > 0 {
-		c.logger.Info("Recovered pending workflows", "count", len(recoveryHandles))
-	} else {
-		c.logger.Debug("No pending workflows to recover")
 	}
 
 	c.logger.Info("DBOS launched", "app_version", c.applicationVersion, "executor_id", c.executorID)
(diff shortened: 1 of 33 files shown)


dbos/internal/models/errors.go
```diff
@@ -36,8 +41,8 @@ func (c ErrorCode) String() string {
 		return "Initialization"
 	case ErrorCodeNonExistentWorkflow:
 		return "NonExistentWorkflow"
-	case ErrorCodeConflictingWorkflow:
-		return "ConflictingWorkflow"
+	case ErrorCodeUnexpectedWorkflow:
+		return "UnexpectedWorkflow"
```
maxdml, line 51: This was really confusing (with the other conflicting workflow ID error used to detect concurrent executions.)

dbos/scheduler.go
```diff
@@ -73,6 +73,33 @@ func jitterCap(sched cron.Schedule, scheduledTime time.Time) time.Duration {
...
+	var v T
+	if len(input.Context) == 0 {
+		return v, nil
+	}
+	if err := json.Unmarshal(input.Context, &v); err != nil {
+		return v, fmt.Errorf("failed to decode schedule context: %w", err)
+	}
+	return v, nil
+}
+
```
maxdml, line 102: We help the user by not decoding json bytes from the database. We help them decode the context bytes in a typed receiver.

dbos/workflow.go
```diff
@@ -1208,19 +1230,26 @@ func (c *dbosContext) RunWorkflow(_ Context, fn WorkflowFunc, input any, opts ..
...
 	// Check if we are within a workflow (and thus a child workflow)
 	parentWorkflowState, ok := c.Value(workflowStateKey).(*workflowState)
 	isChildWorkflow := ok && parentWorkflowState != nil
 
+	// Direct invocations require a launched runtime. Recovery, dequeue, and
+	// child workflow calls are internal paths that may run before Launch completes.
+	if !c.launched.Load() && !params.isRecovery && !params.isDequeue && !isChildWorkflow {
+		c.logger.Error("RunWorkflow called before Launch", "workflow_name", params.WorkflowName)
+		return nil, models.NewInitializationError("DBOS must be launched before running workflows; call Launch first")
+	}
```
maxdml, line 1251: I think this makes more sense: you can't call `RunWorkflow` on a non-launched context anymore, making the distinction clear.

dbos/dbos.go
```diff
@@ -35,10 +35,15 @@ const (
...
+// AppName is required, along with exactly one of DatabaseURL, SystemDBPool,
+// or SQLiteSystemDB.
 type Config struct {
-	AppName                   string          // Application name for identification (required)
-	DatabaseURL               string          // DatabaseURL is the system-database connection string. Exactly one of DatabaseURL, SystemDBPool, or SQLiteSystemDB must be set.
+	AppName string // Application name for identification (required)
+	// DatabaseURL is the system-database connection string. Accepts Postgres URLs or
+	// key=value DSNs (e.g. postgres://user:pass@host:5432/dbname; CockroachDB uses the
+	// same form) and sqlite URLs (sqlite:/path/to.db, sqlite:relative.db, or
+	// sqlite::memory:). Exactly one of DatabaseURL, SystemDBPool, or SQLiteSystemDB must be set.
```
kraftp, line 45: Does sqlite::memory actually work? I remember it not working in Python, it has weird semantics

dbos/workflow.go
kraftp: Is it realistic for someone to create a context before launch and use it after launch?

maxdml: Well, I was abusing this behavior all over the tests for a while. It can happen, I just don't think we should allow it at all, to avoid confusions about what a context is and does

dbos/dbos.go
maxdml: Hmm, I already have a test for testing the sqlite dialect itself on an in-memory sqlite db. I just checked with an actual workflow and it works, but sqlite needs `cache=shared` because otherwise each connection opens a fresh backend. i.e., this only works if you know what you're doing.

PR #407, opened 2026-07-21
Driver registration

maxdml:
To avoid building unrequired dependencies for unused database backends (e.g., modernc for sqlite) this PR adds a db driver registration mechanisms. Pgx and Postgres stay the default.

Here's how it works.
- Add a new subpackage for db drivers, in this first example,  `dbos/driver/sqlite`
- Users import them (here, `_ "github.com/dbos-inc/dbos-transact-golang/dbos/driver/sqlite"`)
- These packages simply import the package they need (here, `modernc.org/sqlite`), which will cause the triggering of an init() function at startup, exposing the driver to the dbos package.
- Without the import, SQLite URLs fail at startup with an error naming the import to add.

Most of the code is unchanged, i.e., the bulk of understanding a driver and using it are still in `dbq.go` and `dialect.go`

### DIFF
--- a/cmd/dbos/main.go
+++ b/cmd/dbos/main.go
@@ -3,7 +3,7 @@ package main
 import (
 	"os"
 
-	_ "github.com/dbos-inc/dbos-transact-golang/dbos/sqlite"
+	_ "github.com/dbos-inc/dbos-transact-golang/dbos/driver/sqlite"
 )
 
 func main() {
--- a/cmd/dbos/main.go
+++ b/cmd/dbos/main.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"os"
+
+	_ "github.com/dbos-inc/dbos-transact-golang/dbos/sqlite"
 )
 
 func main() {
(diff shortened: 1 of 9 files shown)


kraftp (commented):
One consequence of this is that we can't make SQLite the default, which is a bit annoying (one of the reasons it's so nice in Python is that it's a zero-configuration default)

maxdml:
> One consequence of this is that we can't make SQLite the default, which is a bit annoying (one of the reasons it's so nice in Python is that it's a zero-configuration default)

A related question is: do we want a default? Today, Go mandates that at least one of database url / custom pg pool / custom sqlite db is provided. It has the benefit of being simpler and explicit. And we can always have the examples / default apps seed sqlite.

kraftp:
> > One consequence of this is that we can't make SQLite the default, which is a bit annoying (one of the reasons it's so nice in Python is that it's a zero-configuration default)
> 
> A related question is: do we want a default? Today, Go mandates that at least one of database url / custom pg pool / custom sqlite db is provided. It has the benefit of being simpler and explicit. And we can always have the examples / default apps seed sqlite.

That's fair, this still helps demo apps

kraftp (approved):

PR #408, opened 2026-07-21
Debouncer delayed workflows

### LINKED ISSUE #379: Re-implement debouncer with delayed workflows.

maxdml:
Closes #379


dbos/debouncer.go
```diff
@@ -90,13 +116,231 @@ func (o *debouncerOptions) resolveConfigName() string {
...
+			enqueueOpts = append(enqueueOpts, WithEnqueueClassName(params.className))
+		}
+		if params.configName != "" {
+			enqueueOpts = append(enqueueOpts, WithEnqueueConfigName(params.configName))
+		}
+		if workflowTimeout > 0 {
+			enqueueOpts = append(enqueueOpts, WithEnqueueTimeout(workflowTimeout))
+		}
+
+		handle, err := Enqueue[R](c, queueName, params.workflowName, input, enqueueOpts...)
```
kraftp, line 328: If the debouncer is called from a workflow, is this thing checkpointed?

Client enqueue being exposed in the main library, non-checkpointed, seems very dangerous to me.

dbos/debouncer.go
```diff
@@ -312,265 +466,45 @@ func NewDebouncerClient[R any, P any](
...
 	}
 
-	// Configured instance workflows are registered under a name qualified with their config name.
-	if configName := options.resolveConfigName(); configName != "" {
-		workflowName = instanceQualifiedName(workflowName, configName)
+	// The recorded workflow name stays unqualified for configured instances (the config
+	// name is recorded alongside it), but the debounce-key prefix is instance-qualified
+	// so instances of the same method debounce independently.
+	configName := options.resolveConfigName()
+	dedupPrefix := workflowName
```
kraftp, line 489: Does this need class name as well, since we now support that?

dbos/debouncer.go
maxdml: Good catch. Generally speaking, without any notion of client, all of the dbos context methods should detect if they are running within a workflow -- and be checkpointed if so. I'll fix that and make a pass to verify this invariant is tested.

dbos/debouncer.go
maxdml: In Go the class name is not user facing -- we do record one for observability, but it is not used in the registry (keyed by `instance name + function name`). The reason being that, in the other languages, function names are not unique, whereas in Go we have unique identifiers.

dbos/debouncer.go
kraftp: Yeah, good idea. Are there tests calling the debouncer from a workflow?

maxdml:
@kraftp fixed

dbos/workflow.go
```diff
@@ -1882,6 +1873,16 @@ func (c *dbosContext) Enqueue(_ Client, queueName, workflowName string, input an
 		return nil, models.NewInvalidOptionError("a deduplication policy requires a deduplication ID")
 	}
 
+	// Within a workflow, the enqueue is checkpointed as a step; it cannot run inside a step
+	isWithinWorkflow := false
+	wfState, ok := c.Value(workflowStateKey).(*workflowState)
+	if ok && wfState != nil {
+		isWithinWorkflow = true
+		if wfState.isWithinStep {
+			return nil, models.NewStepExecutionError(wfState.workflowID, "DBOS.enqueue", fmt.Errorf("cannot call Enqueue within a step"))
```
kraftp, line 1882: Why not? This also means debounce can't be called from a step, which might be fine

dbos/serialization.go
```diff
@@ -412,6 +414,9 @@ func init() {
 	// Register the scheduled workflow input so gob-serialized schedule firings
 	// round-trip without users having to register it themselves.
 	gob.Register(ScheduledWorkflowInput{})
+	// Register the bounce step's checkpointed output: it is an internal type users
+	// cannot register, and Debounce inside a workflow records it as a step result.
+	gob.Register(&sysdb.DebounceResult{})
```
kraftp, line 431: I still don't fully understand how Go serialization interacts with errors. The debounce flow depends on checking a specific error type, is this safe across serializers (or at least both our regular and portable serializer)?

dbos/workflow.go
maxdml: Generally speaking we don't allow running any of the "special steps" inside a step. Only direct `RunAsStep` and `RunAsTransaction` collapse into the parent durability context.

maxdml: And, yes, that means debounce() should error earlier if we detect it is called from within a step.

dbos/serialization.go
maxdml: The code above is not related to errors: the `DebounceDelayedWorkflow` step returns (and checkpoints) `sysdb.DebounceResult`. Because we do provide a default Gob serializer and this is an internal type, we must gob register it so it can be decoded on the read side, when the user elects to use gob.

We do infer the behavior to adopt, e.g., enqueue fresh workflow, return a polling handle to a debounced workflow, based on the content of the returned `sysdb.DebounceResult`. Not from a specific error type. Am I missing something?

dbos/debouncer.go
maxdml: Yes, there are.

dbos/serialization.go
kraftp: The the code not check for a deduplication error in enqueue anymore?

maxdml: It does, that's part of the `Enqueue` step (when called within a workflow.)

I think your question is whether using a non-default encoder will lose the error type, such that when `Enqueue` is called within a step, the result could be different when replayed.

After further inspection, what I found is:
- For the default serializer, it works (we use gob for errors and it preserves the Code() field, which is essentially what's used for classification)
- For the portable serializer, Go loses the Code() (this is a problem for all paths)

So there is an orthogonal bug which is that Go doesn't capture/parse enough information for the `PortableWorkflowError` type. (Regardless of the path.) I just added the fix to this PR.

A custom serializer goes through the gob path. As a simplification, I only use the user-provided serializer on input/ouputs, not errors. To discuss.

dbos/serialization.go
```diff
@@ -438,6 +497,8 @@ func serializeWorkflowError(logger *slog.Logger, err error, serialization string
 	var errData PortableWorkflowError
 	if pe := (*PortableWorkflowError)(nil); errors.As(err, &pe) {
 		errData = *pe
+	} else if de := (*Error)(nil); errors.As(err, &de) {
```
kraftp, line 500: I don't really understand how this works. Is it tested? How does it handle wrapped errors, or do we not do those anymore?

dbos/serialization.go
```diff
@@ -400,6 +402,18 @@ func (e *PortableWorkflowError) Error() string {
...
+// Is lets errors.Is match a portable error against a DBOS error.
+// serializeWorkflowError records a DBOS *Error's symbolic
+// code (Code.String()) in Name; this compares that to the target sentinel's code,
+// so the match survives a portable (cross-language JSON) round trip.
+func (e *PortableWorkflowError) Is(target error) bool {
+	t, ok := target.(*Error)
+	if !ok || t.Code == 0 {
+		return false
+	}
+	return e.Name == t.Code.String()
```
kraftp, line 414: We're matching on names, and we're giving our errors generic names ("Timeout"), so this will match user errors, no?

dbos/serialization.go
maxdml: We are attempting to cast the incoming error into the dbos package error type. If the assertion matches, we create a new portable error instance fromthe instance of our type.

dbos/serialization.go
maxdml: The error names have to match across SDK. If a user reuse our internal error names, this will collide. We should check whether our names (and codes, for that matter) are in sync across SDKs.

maxdml: Added `TestRecoveryEnqueueWithDedupID`, which exercise recovering a workflow that called `Enqueue` and detected a dedup error, both using the native and the portable serializers. This makes sure that when the workflow is recovered, and reads the outcome of the enqueue step from the database, if there was a dedup error, it is properly decoded as such by the application.

kraftp (approved):

PR #417, opened 2026-07-27
Update workflow status with a final result only when status is PENDING

maxdml:
Update the workflow outcome now results in 5 possible outcomes:
- outcome recorded, the common path
- Error: workflow doesn't exist (wf was garbage collected or deleted)
- Error: workflow was cancelled
- Error: workflow was DLQ-ed
- Park: workflow is DELAYED/ENQUEUED, or already SUCCESS/ERROR. Poll for an outcome.

This does not solve the race if another execution has resumed and the workflow is `PENDING` -- but in this case the outcome should be deterministic and idempotent.

### DIFF
--- a/dbos/workflow.go
+++ b/dbos/workflow.go
@@ -1666,7 +1666,7 @@ func (c *dbosContext) RunWorkflow(_ Context, fn WorkflowFunc, input any, opts ..
 			// visible, a resume→dequeue can re-dispatch this workflow to this executor,
 			// marking it PENDING. But a stale activeID entry would prevent the workflow from running.
 			removeActive()
-			recordErr := sysdb.Retry(c, func() error {
+			recordAction, recordErr := sysdb.RetryWithResult(c, func() (sysdb.OutcomeWriteAction, error) {
 				return c.systemDB.UpdateWorkflowOutcome(uncancellableCtx, sysdb.UpdateWorkflowOutcomeDBInput{
 					WorkflowID: workflowID,
 					Status:     status,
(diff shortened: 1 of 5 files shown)


kraftp (commented):
Can we simplify this to always "park" the workflow? Then the `getResult` called when you "park" should already correctly handle all the other outcomes

maxdml:
> Can we simplify this to always "park" the workflow? Then the `getResult` called when you "park" should already correctly handle all the other outcomes

There is a trade off between having all executions fall back to the eventual outcome, and accurately reporting what a given execution encountered. Given how complex these races can be to reason about, I'd lean toward being explicit (not withstanding the performance improvements.) The alternative being to rely on logging only.

kraftp:
> > Can we simplify this to always "park" the workflow? Then the `getResult` called when you "park" should already correctly handle all the other outcomes
> 
> There is a trade off between having all executions fall back to the eventual outcome, and accurately reporting what a given execution encountered. Given how complex these races can be to reason about, I'd lean toward being explicit (not withstanding the performance improvements.) The alternative being to rely on logging only.

This makes it substantially more complex. Log and move on, I think, having a unified error-handling path is a big improvement.

dbos/workflow.go
```diff
@@ -1675,21 +1675,20 @@ func (c *dbosContext) RunWorkflow(_ Context, fn WorkflowFunc, input any, opts ..
...
-					close(outcomeChan)
-					return
-				}
 				c.logger.Error("Error recording workflow outcome", "workflow_id", workflowID, "error", recordErr)
 				outcomeChan <- workflowOutcome[any]{result: nil, err: recordErr}
 				close(outcomeChan)
 				return
 			}
+			if !recorded {
+				// The row was not PENDING: this run no longer owns the workflow's
```
kraftp, line 1689: This is much simpler, we can continue to have a separate case for when the workflow was deleted entirely

kraftp (approved):

dbos/workflow.go
maxdml: Well that would require adding back the 2nd read...

PR #420, opened 2026-07-30
Restrict nested steps

maxdml:
Forbid running `RunAsTransaction`/`GetResult`/`SetEvent`/`CloseStream` inside a step (today falls back to the parent durability context), preferring clarity over difficult semantics. Read operations (e.g., `ListWorkflows`) and `WriteStream` are still allowed.

### DIFF
--- a/dbos/errors_test.go
+++ b/dbos/errors_test.go
@@ -1,0 +1,90 @@
+package dbos
+
+import (
+	"context"
+	"errors"
+	"path/filepath"
+	"testing"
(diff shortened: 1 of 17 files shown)


dbos/errors_test.go
maxdml, line 1: A missing file in my local.

kraftp (approved):
`SetEvent` should definitely be allowed for the same reason as `WriteStream`. Restricting the rest is fine.

PR #433, opened 2026-08-13
Always enqueue recovered workflows

### LINKED ISSUE #406: Executor-filtered recovery + always queue based recovery
- When clearing queue assignment for recovering queued tasks, filter by the list of executor IDs to recover (this prevents recovering workflows that are already being recovered)
- Non-queued workflow to be recovered on the internal queue

maxdml:
Also:
* Batch recovery enqueues
* Filter recovered workflows by the list of executors to recover
* Store non-queued workflows queue name as null instead of an empty string (made backward compatible with `NULLIF`)

Closes #406

### DIFF
--- a/dbos/recovery.go
+++ b/dbos/recovery.go
@@ -1,90 +1,22 @@
 package dbos
 
 import (
-	"errors"
-
+	"github.com/dbos-inc/dbos-transact-golang/dbos/internal/models"
 	"github.com/dbos-inc/dbos-transact-golang/dbos/internal/sysdb"
 )
 
+// recoverPendingWorkflows re-enqueues pending workflows owned by the specified executors and returns handles to them.
 func recoverPendingWorkflows(ctx *dbosContext, executorIDs []string) ([]WorkflowHandle[any], error) {
-	workflowHandles := make([]WorkflowHandle[any], 0)
-	// List pending workflows for the executors
-	pendingWorkflows, err := sysdb.RetryWithResult(ctx, func() ([]WorkflowStatus, error) {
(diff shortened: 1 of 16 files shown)


kraftp (commented):
One change we made in Python that probably should accompany this is validating that a deduplication key must be tied to a queue name. Otherwise a dedup ID could be added to the internal queue, which is bad

dbos/recovery.go
```diff
@@ -1,90 +1,22 @@
...
-		registeredWorkflow, ok := registeredWorkflowAny.(WorkflowRegistryEntry)
-		if !ok {
-			ctx.logger.Error("invalid workflow registry entry type", "workflow_id", workflow.ID, "name", workflow.Name)
-			continue
-		}
-
-		// Convert workflow parameters to options.
-		// Auth identity is re-attached so child workflows spawned during
-		// recovery inherit the same identity as the original run.
-		opts := []WorkflowOption{
```
kraftp, line 68: This went away and didn't come back?

maxdml: They are already in the DB. The problem actually is that the dequeue path doesn't add them https://github.com/dbos-inc/dbos-transact-golang/blob/c32259f56f50e87c37aa3e67bf07e6f56eb92891/dbos/queue.go#L696

maxdml:
> One change we made in Python that probably should accompany this is validating that a deduplication key must be tied to a queue name. Otherwise a dedup ID could be added to the internal queue, which is bad

fixed

kraftp (approved):

PR #444, opened 2026-08-17
Add shared-base migrations 100-107 for application_name

maxdml:
Schema migrations in preparation for sharing system database among applications

### DIFF
--- a/dbos/migrations.go
+++ b/dbos/migrations.go
@@ -14,10 +14,12 @@ import (
 // semicolon-terminated statements and "--" comment lines suitable for
 // execution with psql. It never connects to a database.
 //
-// from is a 1-based migration number: pass 1 for the full fresh-database
+// from is a migration version number: pass 1 for the full fresh-database
 // script (including schema creation, the dbos_migrations table, and the
 // initial version row), or a later number for migrations from through the
-// latest only. Each migration is followed by its version bookkeeping,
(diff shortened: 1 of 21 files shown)


dbos/migrations.go
```diff
@@ -14,10 +14,12 @@ import (
 // semicolon-terminated statements and "--" comment lines suitable for
 // execution with psql. It never connects to a database.
 //
-// from is a 1-based migration number: pass 1 for the full fresh-database
+// from is a migration version number: pass 1 for the full fresh-database
```
kraftp, line 17: Two quick sanity checks:

- We're not doing blank transactions or printing garbage for the blank migrations?
- We're not emitting anything for the blank migrations from the CLI print command?

kraftp (approved):

dbos/migrations.go
maxdml: > We're not doing blank transactions or printing garbage for the blank migrations?

No, when skipping a migration, we just bump the ID, we don't do a blank transaction. See how https://github.com/dbos-inc/dbos-transact-golang/pull/444/changes#diff-6b92e5210850374f06c5d1c6ddca64cc96ac6a9355134d2048a84c8393af3306R550-R559 is appending the new numbers (versions and files) to the list that is iterated


> We're not emitting anything for the blank migrations from the CLI print command?

nope

Screenshot 2026-08-17 at 17 10 19

PR #459, opened 2026-08-30
add: SendBulk API

### LINKED ISSUE #435: Add bulk send
Send multiple messages to a workflow in one call/DB round-trip. Parity: Python `send_bulk`, Java.

toastsandwich:
closes #435 

disclaimer: grok showed a new way to insertNotifications, check that once.  If you want the previous one I will just switch to it : )


dbos/internal/sysdb/system_database.go
```diff
@@ -4109,49 +4109,90 @@ func (s *SysDB) pollEvents(ctx context.Context) {
...
+		if err := s.insertNotificationChunk(ctx, exec, input.Messages[start:end], baseMs+int64(start)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *SysDB) insertNotificationChunk(ctx context.Context, exec Querier, msgs []WorkflowSendRow, baseMs int64) error {
+	valueRows := make([]string, 0, len(msgs))
+	args := make([]any, 0, len(msgs)*sendInsertCols)
```
maxdml, line 4177: Let's make sendInsertCols an actual slice of the columns (`destination_uuid, topic, message, serialization, message_uuid, created_at_epoch_ms`), declared in this method, and take its length.

dbos/workflow.go
```diff
@@ -3143,22 +3145,83 @@ func WithSendTransaction(tx any) SendOption {
...
+	}}, sendStepName, options)
+}
+
+func (c *dbosContext) SendBulk(_ Client, messages []SendMessage, opts ...SendOption) error {
+	options := &sendOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
+	if options.idempotencyKey != "" {
+		return models.NewInvalidOptionError("WithIdempotencyKey is per-message; set SendMessage.IdempotencyKey")
```
maxdml, line 3182: Let's be even more explicit with that error message: 
```suggestion
		return models.NewInvalidOptionError("WithIdempotencyKey is per-message; set SendMessage.IdempotencyKey when using SendBulk")
```

dbos/internal/sysdb/system_database.go
```diff
@@ -4109,49 +4109,90 @@ func (s *SysDB) pollEvents(ctx context.Context) {
...
+	}
+
+	baseMs := time.Now().UnixMilli()
+	for start := 0; start < len(input.Messages); start += sendInsertChunk {
+		end := min(start+sendInsertChunk, len(input.Messages))
+		if err := s.insertNotificationChunk(ctx, exec, input.Messages[start:end], baseMs+int64(start)); err != nil {
+			return err
+		}
+	}
+	return nil
```
maxdml, line 4172: Batches need to be inserted atomically, i.e., within the same transaction. On the common path where input.Tx is null, the current code does one insert per batch. Let's fix it and add a test that has at least 2 batches and simulates a failure after the first one: no rows should have been inserted.

dbos/internal/sysdb/system_database.go
```diff
@@ -4109,49 +4109,90 @@ func (s *SysDB) pollEvents(ctx context.Context) {
 
 const NullTopic = "__null__topic__"
 
-type WorkflowSendInput struct {
+const (
+	sendInsertCols  = 6
+	sendInsertChunk = 1000 // 1000*6 bind params, under Postgres 65535
```
maxdml, line 4130: Let's make the batch size configurable in `WorkflowSendInput` and default to:

`const _DEFAULT_BULK_SEND_BATCH_SIZE = 5000`

(Let's set 5k so it fits both sqlite limits and postgres)

The public interface in workflow.go can accept that has a new send option, `WithSendBulkBatchSize` (if used in raw `Send`, emit a warning saying this is a bulk-only option)

maxdml (approved):

Issue #464, opened 2026-09-10
Cloned contexts panic on launch


jnschaeffer:
When using DBOS contexts cloned from other DBOS contexts (e.g., when using `WithValue` or similar functions), the resulting context will panic with a nil pointer dereference error after calling `Launch`.

Example program tested against latest main (8279c97):

```go
package main

import (
	"context"
	"log/slog"
	"os"
	"os/signal"
	"time"

	"github.com/dbos-inc/dbos-transact-golang/dbos"
)

func main() {
	logger := slog.Default()
	logger.Info("starting DBOS")
	c := make(chan os.Signal, 1)
	signal.Notify(c, os.Interrupt)
	dbosContext, err := dbos.NewContext(context.Background(), dbos.Config{
		AppName:            "dbos-starter",
		ApplicationVersion: "0.1.0",
		DatabaseURL:        os.Getenv("DBOS_SYSTEM_DATABASE_URL"),
		Logger:             logger,
	})
	if err != nil {
		slog.Error("error creating DBOS context", "error", err.Error())
		os.Exit(1)
	}
	newCtx := dbos.WithValue(dbosContext, "foo", "bar")
	if err := dbos.Launch(newCtx); err != nil {
		slog.Error("error creating DBOS context", "error", err.Error())
		os.Exit(1)
	}
	defer dbos.Shutdown(newCtx, 5*time.Second)
	<-c
}
```

Output:
```
$ ./dbos-bug-repro 
2026/09/09 21:16:33 INFO starting DBOS
2026/09/09 21:16:33 INFO Initializing DBOS context app_name=dbos-starter dbos_version=(devel)
2026/09/09 21:16:33 INFO Connecting to system database database_url=postgresql://postgres:***@localhost:5432/postgres schema=dbos
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x488890]

goroutine 1 [running]:
github.com/dbos-inc/dbos-transact-golang/dbos.(*dbosContext).Shutdown(0x654b8c5eb7a0, {0x596e8?, 0x654b8c7df7b8?}, 0x6fc23ac00)
        [REDACTED]/dbos-transact-golang/dbos/dbos.go:915 +0xc0
github.com/dbos-inc/dbos-transact-golang/dbos.(*dbosContext).Launch.func1()
        [REDACTED]/dbos-transact-golang/dbos/dbos.go:807 +0x44
[...]

PR #465, opened 2026-09-10
Copy workflowScheduler and ctxCancelFunc when cloning contexts

jnschaeffer:
This PR updates `(*dbosContext).clone` to include `workflowScheduler` and `ctxCancelFunc` from the original context. This fixes two issues. The first is that `(*dbosContext).Launch` panics when using a cloned context, as the context attempts to call `Start` on a nil workflow scheduler. The second issue is that `(*dbosContext).Shutdown` panics when calling a nil `ctxCancelFunc`.

My initial intent was to address #464 by merely copying `workflowScheduler` over from the original context during cloning, but on shutdown the cloned context also panicked because it was missing `ctxCancelFunc`. As a result, this PR adds both fields to the cloned context. There may be additional fields that need to be copied over, or tests that need to be written to exercise launching of cloned contexts.

### DIFF
--- a/dbos/dbos.go
+++ b/dbos/dbos.go
@@ -403,9 +403,11 @@ func (c *dbosContext) Value(key any) any {
 func (c *dbosContext) clone(ctx context.Context) *dbosContext {
 	childCtx := &dbosContext{
 		ctx:                     ctx,
+		ctxCancelFunc:           c.ctxCancelFunc,
 		config:                  c.config,
 		logger:                  c.logger,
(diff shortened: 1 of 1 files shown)


maxdml:
Thanks @jnschaeffer . What's the use case for calling `Launch`/`Shutdown` on a cloned context? This isn't supported. Happy to consider supporting it if there's a real use case.

maxdml:
@jnschaeffer you can set the parameters you need in the `context.Context `passed to `NewContext`

PR #468, opened 2026-09-11
Close gzip writer on export error path and query rows after each fork-test iteration

kojah:
Two small resource cleanup fixes:

- `handleExportWorkflowRequest` in `conductor.go` leaves the gzip writer unclosed when `gz.Write` fails. Added `gz.Close()` on that error path.
- `TestForkWorkflow` deferred `rows.Close()` to test end, so open row sets accumulated across loop iterations. Moved the close to immediately after each iteration finishes scanning.

**Testing**

`go build ./dbos/...`, `go vet ./dbos/...`, and `go test -run '^$' ./dbos/...` all pass.

I found these issues through a static analysis tool I'm developing called [gohawk](https://github.com/kojah/gohawk). If you're interested, let me know and I can open a separate PR to add gohawk to your linter configuration :) You can view more comprehensive information about the project on the documentation website: [https://gohawk.dev](https://gohawk.dev)


dbos/conductor.go
```diff
@@ -1307,6 +1307,7 @@ func (c *conductor) handleExportWorkflowRequest(data []byte, requestID string) e
 			var buf bytes.Buffer
 			gz := gzip.NewWriter(&buf)
 			if _, err := gz.Write(jsonData); err != nil {
+				_ = gz.Close()
```
maxdml, line 1310: Let's handle errors

dbos/client_test.go
```diff
@@ -1010,6 +1009,7 @@ func TestForkWorkflow(t *testing.T) {
 				require.NoError(t, err, "failed to unmarshal value")
 				actualEventTuples = append(actualEventTuples, eventTuple{functionID, key, value})
 			}
+			rows.Close()
```
maxdml: This change is not safe. While defer calls accumulate, they end up in a no-op if the rows set is already close (which it does automatically when the connection is released).

But if we do this and the test fails mid-execution, then rows won't be closed and goleak will fail.

maxdml (approved):

PR #470, opened 2026-09-11
Workflow payload tables

maxdml:


kraftp (commented):
One quick check--is Go consistent (as I believe Python and TS are now) at using database-generated timestamps for all row creation timestamps?

kraftp (approved):

maxdml:
> One quick check--is Go consistent (as I believe Python and TS are now) at using database-generated timestamps for all row creation timestamps?

Not fully there yet. This PR and a previous one go in that direction. I'll align in a next PR

